PR #13, opened 2022-09-27
Switch to MapLibreGL and use ArcGIS tileset with ESRI source support

### LINKED ISSUE #8: Use ESRI Mapbox instead of Mapbox tiles
https://frontiersi.github.io/mapbox-gl-esri-sources/#/installation

jxjj:
## Changes

- Change from MapboxGL to the open source MapLibreGL.
- Use ArcGIS API key for maps rather than Mapbox Access token
- Support esri sources (I think? ... it'd be nice to test this with something).

maplibre-map

## Note about `Type Error: Type instantiation is excessively deep and possibly infinite.`

MapLibre keeps complaining: `Type instantiation is excessively deep and possibly infinite.` when creating a new Map or Map Marker. I think it may have something to do with the map style types, but I'm not sure. Ultimately, I gave up debugging and just wrapped the problematic parts with `@ts-ignore`.

While debugging, I tried bumping deps... so that's also in this PR.

Anyhoo: @cmcfadden – Let me know if this aligns with the direction we discussed last week, or if I'm missing something. I'm not entirely clear if/how Esri and ArcGIS are used across elevator instances beyond base tile maps.

Feedback welcome.

Closes #8


cmcfadden:
It's really just base maps that we use from esri AFAIK. 

Are we able to toggle between map/satellite? Is that alert prompt a one time thing or does that come up on every load?

Interesting that you had to provide an API key, as we didn't with the leaflet package. Do they have any published limits or other things we should be aware of as part of getting that API key?

cmcfadden:
ah apparently they added Api stuff to V3 of the leaflet plugin too. Hmm. Looks like 2m tiles for free?

jxjj:
> Are we able to toggle between map/satellite?

Done!

> Is that alert prompt a one time thing or does that come up on every load?

Once per load. It's the default attribution info required by ESRI. On larger screens, it's fine: just a transparent overlay at the bottom. When the map gets smaller and the overlay changes to a white background that's collapsable. (Not my design... just a standard part of MapLibre.) On the bright side, once we add in all the cookie consent klaxons, they won't even notice it!

jxjj:
Updated gif:
Screen Recording 2022-09-27 at 8 32 17 PM

cmcfadden:
Cool!

PR #14, opened 2022-09-28
Detect when asset has metadata only, and render in a Meta Only Page

jxjj:
This adds a story and view for metadata only assets. When an asset is loaded, if no `firstFileHandlerId` is present, the asset it determined to be "meta-data only", and renders the data without a viewer.

The view is minimal. (Once the share and download tools are implemented, we can add those to the bottom or side).

Large screen:
Screen Shot 2022-09-28 at 6 05 21 PM

Small screen:
Screen Shot 2022-09-28 at 6 05 32 PM

### DIFF
--- a/src/components/Drawer/Drawer.vue
+++ b/src/components/Drawer/Drawer.vue
@@ -83,19 +83,4 @@ defineEmits<{
 ::-webkit-scrollbar-thumb:hover {
   background: #555;
 }
-
-.ticker {
-  animation: 8s ticker infinite linear;
-}
-
-@keyframes ticker {
-  0% {
-    transform: translate3d(0, 0, 0);
-    visibility: visible;
-  }
-
-  100% {
-    transform: translate3d(-100%, 0, 0);
-  }
-}
 </style>
--- a/src/pages/AssetViewPage.stories.ts
+++ b/src/pages/AssetViewPage.stories.ts
@@ -3,6 +3,9 @@ import AssetViewPage from "@/pages/AssetViewPage.vue";
 
 export default {
   component: AssetViewPage,
+  parameters: {
+    layout: "fullscreen",
+  },
 } as Meta<typeof AssetViewPage>;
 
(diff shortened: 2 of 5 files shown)


cmcfadden:
There won't be any share/download tools in this case, since there's no digital asset. 

Seems good - we can probably go wider on the big screen device, but that can be dealt with down the line.

cmcfadden:
Actually, one thing we should tackle here - it would be possible to build an asset with no firstFileObject but where a user could still cause an asset to load. It looks like this wouldn't flip to the other view in that case? Can the page type just be dependent on whether or not we have a fileObject specified in the store?

cmcfadden:
https://dev.elevator.umn.edu/defaultinstance/asset/viewAsset/6334d6997591f67a5976fbd1

jxjj:
Thanks - I'm getting those 401 errors for that page when trying to fetch the json.

Hmmm... I think the `watch` would pick up the `assetId` change, and recalc whether or not the firstFileHandlerId is defined when it gets it from the store (or fetches).

I changed the logic to be computed from `activeFileHandlerId` in the pinia store, so it's a bit clearer – thanks for the suggestion. 

I _think_ this is equivalent to the previous functionality, though, so maybe I'm misunderstanding something about this case?

cmcfadden:
Ugh sorry, try now

jxjj:
Cool, thanks!

Here's what the initial page render looks like with that asset:
Screen Shot 2022-09-28 at 7 09 32 PM

And after clicking the related asset, it shifts to:
Screen Shot 2022-09-28 at 7 09 47 PM

Expected functionality or not quite right?

jxjj:
Recording if it helps:

cmcfadden:
Sounds right to me – just to clarify, it happens when you click on the thumbnail, right?

cmcfadden:
Woops sorry, didn’t see the recording!

looks perfect!

PR #29, opened 2022-11-01
add loading lazy to images

jxjj:
This adds `loading="lazy"` to `img`. [Browser support](https://caniuse.com/loading-lazy-attr) is pretty good for images, so I think we're pretty good without a polyfill.

This also removes the "Old View" and "JSON" buttons that I added to the bottom of the drawers assist with development.


cmcfadden:
I'm not sure this is doing anything? Looking at network inspector while loading [this asset](https://dev.elevator.umn.edu/defaultinstance/asset/viewAsset/5c06f4dc10c07d78e71d9624) shows all the images being pulled down?

jxjj:
Thanks for testing.

Hmm... I'm seeing images download once assets scroll into view:

Screenshot 2022-11-01 at 20 11 11

Screenshot 2022-11-01 at 20 07 29

I'll see if I can reproduce what you're seeing.

jxjj:
Yep, I'm seeing the same thing you are in Safari.

Possible workaround: add a pop-up with a download link to a real browser? 😉

cmcfadden:
Fired.

PR #30, opened 2022-11-03
Optimize icon font importing

### LINKED ISSUE #28: optimize material symbols
pretty large download (2.1MB).
We don't need all of this

jxjj:
This: 
- replaces Google Material Symbols with icon components in `src/icons`
- swaps the Google fonts import of Works Sans with a locally installed version

Also: in the main Elevator repo, I removed the links to the stylesheets for work sans and the material symbols icon font from vueTemplate.php.

## Background
Material Symbols is a big download (~2MB), and on slow connections users can get a flash of the icon words (`arrow_forward`) in place of the icons until the icon font is downloaded.

Since we only use a very small subset of icons, this PR swaps out Material Symbols with a component for each icon that we use. 

I investigated using some sort of [font glyph subsetting like FontAwesome Pro](https://fontawesome.com/docs/web/dig-deeper/subsetter) has, but it seemed like overkill for our needs. Ultimately, I decided to keep it simple and settled on just using SVG embedded in component. This also us to mix and match icons between different sets or manipulate with CSS/JS  if needed.

Finally, I also changed to [HeroIcons](https://heroicons.com/), since the art was similar in style but has a nicer one-click interface for getting the svg.

Closes #28


cmcfadden (commented):
looks good to me

cmcfadden (approved):

PR #41, opened 2022-11-08
Add support for theming

### LINKED ISSUE #22: Reduce contrast between collapsed drawers
less Cruella de Vil, more Billie Eilish?

### LINKED ISSUE #31: Make download modal smaller?
We'll never have more than 4 options in the download list so we can probably optimize that modal size a bit?

### LINKED ISSUE #39: Reduce gap between fields?
I feel like we can get away with less vertical padding between fields, so that more content is above the fold. Doing a gap-4 instead of gap-8 still feels like plenty of padding to me. Thoughts?

**Gap 8:**
Screenshot 2022-11-05 at 12 17 08 PM

**Gap 4:**
Screenshot 2022-11-05 at 12 16 10 PM

### LINKED ISSUE #40: Checkbox widget broken with icon updates?
Not sure if this is a caching/build issue or if we need to update the icon in use:

Screenshot 2022-11-05 at 12 36 55 PM

jxjj:
This PR adds support for creating themes for Elevator UI, extracting a number of design decisions to CSS variables for easier modification. 
- Currently, there are two themes, a neutral light theme and a dark theme. 
- The "Auto" theme will detect system preferences if one exists, using dark mode if your system is set to dark mode. And light mode otherwise. It watches the system preferences, so if the pref changes, the page theme should update without reload.
- The light theme gives both panel headers the same background (rather than white and black), which fixes #22.
- Default WidgetList gap is set to 1.5rem, and configurable with a CSS variable so we can create a theme with less "whitespace" if desired, which resolves #39.

The dark theme was created to mostly to identify which CSS variables were missing. I was surprised how fast it went to create a new theme. Maybe a few hours from design to implementation and  most of that time was fiddling with color choices or adding new CSS vars for props that I had hardcoded in the light theme.

Also:
- fixed #40 – a checked circle icon and unchecked circle icon have been added. This broke after Material Symbols CSS was removed. (Storybook still had a link to the material symbols CSS in it's head, so that was also removed).
- fixed #31 – modal window's height is more appropriately sized to the content.

Some caveats:
- The "Menu" button doesn't do anything yet. It's just there as a placeholder for now.
- The "Theme" dropdown is also in a temp location. Once we have a better sense of menu content and search stuff, it may make more sense to move it or reformat.

### Asset View
Screenshot 2022-11-07 at 22 40 05@2x
Screenshot 2022-11-07 at 22 41 25@2x

### Meta Data Only
Screenshot 2022-11-07 at 22 43 18@2x
Screenshot 2022-11-07 at 22 43 57@2x

### Modal
Screenshot 2022-11-07 at 22 44 41@2x
Screenshot 2022-11-07 at 22 45 04@2x
Screenshot 2022-11-07 at 22 45 25@2x


jxjj:
Currently up on dev: https://dev.elevator.umn.edu/dcl/asset/viewAsset/56a3bb007d58ae8a488b4657

(cache clear needed to see update)

cmcfadden (approved):
This looks super great. Not sure how I feel about giving into dark mode users though. Maybe every 15 minutes it should just flip to light mode, to remind them what they're missing?

(seriously, super cool)

jxjj:
Thanks!

PR #44, opened 2022-11-10
Wait for object viewer load before loading MoreLikeThis content

### LINKED ISSUE #17: Delay looking for related items until page has loaded

jxjj:
Waits for load event from iframe before starting load of MoreLikeThis content.
Screenshot 2022-11-10 at 16 17 24

 Adds a generic thumbnail image for related assets if the `primaryHandlerId` is `null`
Screenshot 2022-11-10 at 16 27 26@2x

Other:
- fixed Storybook css load issue (from moving app.css in another PR)
- added npm script `dev:remotedev` which will do a build watch in development mode, which is helpful for debugging with vue dev tools (normally disabled in production).
- fixed an issue where `getScrollParent` with throw in Storybook (it would try to getComputedStyles of window)
- added theme support for search results card

Closes #17


src/components/ObjectViewer/ObjectViewer.vue
```diff
@@ -20,10 +21,24 @@
...
 defineProps<{
   fileHandlerId: string | null;
 }>();
+
+const emit = defineEmits<{
+  (eventName: "objectViewLoad");
+}>();
+
+onMounted(() => {
+  console.log("ObjectViewer mounted");
```
cmcfadden, line 35: Do we want to leave these console logs? I'm fine with having some standard debug logs in the codebase if we think it's useful, but we should do some standardize prefixing if they're gonna be in prod

src/pages/AssetViewPage/AssetView.vue
```diff
@@ -45,33 +48,37 @@
...
+import { ref, defineAsyncComponent } from "vue";
 import { useAssetStore } from "@/stores/assetStore";
 import ObjectViewer from "@/components/ObjectViewer/ObjectViewer.vue";
 import ObjectDetailsDrawer from "@/components/ObjectDetailsDrawer/ObjectDetailsDrawer.vue";
 import AssetDetailsDrawer from "@/components/AssetDetailsDrawer/AssetDetailsDrawer.vue";
-import { useMoreLikeThis } from "@/helpers/useMoreLikeThis";
 import { useMediaQuery } from "@vueuse/core";
 
-const props = defineProps<{
+const MoreLikeThis = defineAsyncComponent(
```
cmcfadden, line 65: Why is this one an async component?

cmcfadden (commented):
Any idea why this PR has so much class name churn? Linter update to start alphabetizing them?

Glad this works, seems like a good solution. I'd like to keep the "more like this" button revealing a modal, instead of drawing the results all the time. Thoughts?

jxjj:
Maybe a few things:

- I installed stylelint and prettier tailwind plugin locally which orders
classes in a standard way (I was annoyed that my linter missed a few typos css errors and typos I made). Fix on save reordered stuff. 

- I removed padding from the default accordion to solve an issue with weird padding on accordions with images. 

- handling long titles with ellipses in ThumbnailGeneric (for null asset handlers) required more class shuffling 

- also renamed a few components along the way.

jxjj:
Why a modal?

src/components/ObjectViewer/ObjectViewer.vue
jxjj: Nope. We can remove.

cmcfadden:
Ah that makes sense on classes. I was mostly just noticing cases where the names were rearranged to be alphabetical, which seemed automatic. All good. 

Regarding the "more like this", here's my thinking

- For almost all elevator sites, even if they have >0 results, they're not especially useful. That could change in the future if we rethinking the search system
- Even on a site like the DCL where they are potentially useful, most users (can check analytics but usually >70%) are landing on a page through a direct link to an asset, so they're not really interested in just browsing the system. Having the more like this content displayed all the time is just noise that distracts from their goal.
- The modal gives us more space to show the related items so that for the rare user who stands to benefit, they can get a better sense of the context and the relationship

Since we have the data, we could badge the button with the results count to help make it clear what they can expect. And I'm not wed to a modal specifically, moreso just to not having the related assets drawing on the main asset layer.

jxjj:
Thanks!

Let's talk more next week when you're back, but I'll jot down my current thinking:

1.  Analytics will tell us how users are using the *current* site, but not their intent. Users might not be browsing because it's not easy to browse, or the results surfaced aren't helpful. It's muddy. In my casual usage of the current site, More Like This required an extra click and was often empty. Eventually, I just stopped checking it. User interviews would give better info about intent, if we're interested in learning more.

2. I don't think it's distracting. It's at the very bottom of the asset panel. Users will need to scroll to find it (and know which panel to find it on, which is itself a problem). It won't show up on most pages (again, this could be confusing to users looking for it.) It might be confusing where we hid it, but it's not distracting where we put it.

3. I agree that users may want more space to view results if they're interested. In which case, I would consider limiting the results to say 3 "more like this" and then have a "See More" link which could take them to a Search page with all the results. 

4. Load order fiddling smells like premature optimization. The solutions here may or may not be applicable to other pages, so it might be better to wait on further optimization until rest of the ui and backend work is done.

5. I personally love the idea of More Like This. If it can give good results, I think we should highlight it. Even if not clicked, it gives users a sense of scale of the data available to them. It's true that it can lead users off their original path, but that's the point. Meaningful research, teaching, and learning isn't linear, and those serendipitous moments off-trail of can make all the difference.

6. Sometimes you need to kill your darlings. If you're right and most users aren't interested in More Like This, then we should let it go. Optimizing around an unused feature and then burying it in the UI isn't really doing users any favors, and it has an opportunity cost for us.

Mull things over and we can decide how to proceed. I leave this PR as is for now.

src/pages/AssetViewPage/AssetView.vue
jxjj: Async component allows us to:
1. Skip loading code that's not needed.
2. Add a delay if we want.
3. Add an intersection observer and lazy load if we want.

I didn't know if the load performance would be satisfactory to you, so I made this as a async component in case I needed to tweak further. 

It's trivial to switch to a regular import if you have a preference. I don't really.

cmcfadden: Maybe I'm misunderstanding how Vue handles this - will it only load the async component if the <morelikethis> element is actually being rendered (IE, if the v-if resolves to true)? That'd be super cool. I was assuming this would cause the MoreLikeThis to be loaded on every AssetView regardless?

Slightly larger bundle size / longer initial load is overall preferable vs the latency and connection limiting on the server side. Can definitely re-examine this when we can move the servers to HTTP/2, which reduces that overhead.

cmcfadden:
FYI you should get better real-ish results on /dcl/ now, using assets like this for example:
https://dev.elevator.umn.edu/dcl/asset/viewAsset/56a3c12a7d58ae1f4b8b4dd3

PR #47, opened 2022-11-18
Limit number of  inline results of More Like This section

jxjj:
This limits the number of results we show in the More Like This section of the asset panel to an instance configurable number (default 3). If there are more results, users can click "Show More" which will open a modal of all results.

If an instance is configured to with `config.moreLikeThis.maxInlineResults` to be 0, no inline results will be shown and instead the users will just have a button to open More Like This results as a modal.

Screenshot 2022-11-18 at 10 27 12@2x

Screenshot 2022-11-18 at 10 27 32@2x

Screenshot 2022-11-18 at 10 30 38@2x

As discussed, we can hone the design elements for the More Like This section once we work out the components on the search results page. (e.g. Maybe a flatter row style component might look better here than a card.)

### DIFF
--- a/src/components/Modal/Modal.vue
+++ b/src/components/Modal/Modal.vue
@@ -20,7 +20,11 @@
         <header
           class="modal-contents__header flex justify-between items-start p-4 md:p-8"
         >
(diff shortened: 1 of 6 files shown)


cmcfadden (approved):
👍

PR #49, opened 2022-11-29
Add support for LTI (Canvas) and CMS (WordPress) embedding 

jxjj:
This adds support for Canvas and WordPress embedding.

When a page is opened as an embedded item as indicated by values in sessionStorage, an "Add to ..." button will appear above (not overlapping) the object viewer . When a user clicks the "Add button", an interstitial confirm message will appear (if set), and the asset will be embedded.

## ✨  Added
- AddToPluginButton component, which contains most of the embed logic
- ConfirmModal component, to support the interstitial dialog
- icons for showing various states of embed process (loading, error, success)
- `useElevatorSessionStorage` - a small wrapper for getting all the elevator specific session variables we care about. Also this includes a `clear()` method for resetting these variables on destroy.

##  ⬆️ Updated
- `AssetDetailsDrawer` - only show MoreLikeThis when there's an asset id.
- `ObjectViewer` - changed layout to make space for the AddToPluginButton
- `routes` - Added a test route for working on lti component. Also refactored so that test routes only are included if we're in `development` mode.
- `api` - add calls to get interstitial message, and post the lti payload to Elevator

##  🐛 Fixed
- MetaDataOnlyView - add missing `MoreLikeThis` section for components that only have meta data.

## 🍿 Demos

Wordpress:
Screenshot 2022-11-29 at 13 10 02

Canvas:
Screenshot 2022-11-29 at 13 11 42


cmcfadden (approved):
looks great

PR #60, opened 2022-12-13
Add Search Results page

jxjj:
This is an initial version of the search results page.  Not quite done, but ready for feedback on design and functionality.

There's an issue with vue rerendering all results when more are added to the array. This makes infinite scroll disorienting, so I switched to Load More button,

Screenshot 2022-12-13 at 01 07 18@2x


src/types/index.ts
```diff
@@ -305,13 +305,27 @@ export interface SearchResultMatch {
   primaryHandlerThumbnail2x?: string; //URI
 }
 
-export interface SearchResponse {
+export interface SearchEntry {
+  collection?: number[];
+  searchDate?: DateTime;
+  searchText?: string;
+  matchType?: string; // 'phrase_prefix' ?
+  showHidden?: "0" | "1";
```
cmcfadden, line 313: does our api do this??? ughhhh I'm sorry

cmcfadden:
Looks good! Couple other things, may/may not be for this PR: 

* Feels like there's still something odd with infinite scroll, at least on safari? The assets kind of jump around when the load happens in a way that's disorienting. 
* Badging the asset count for items with >1 digital asset
* Truncating long items to a reasonable height
* The cover-fit rendering seems good for images but kind of weird for the placeholder icons (like the blank document icon). Maybe our best bet is to make a set of those icons in 16:9 for the Vue interface?

jxjj:
Helpful – thanks!

> Truncating long items to a reasonable height

Any issue with limiting the number of "Tuples" displayed below the title? What's a good number? 3?

cmcfadden:
With the old layout, we used an overflow-y:scroll in the box rather than truncate. I don't *think* we want to do that, but some sorta max-height makes more sense than limiting the number of tuples. The tuples displayed are specifically selected by the instance admins for the search results, so they've said "these are the things that are relevant in a search result" .. don't think we want to filter more than that. Mostly just trying to tackle a worst case like the attached one. Maybe set a max-height for an individual tuple? Thoughts?

Screenshot 2022-12-13 at 9 53 30 AM

jxjj:
Whoa. That's a big one. Yeah, max-height is the way to go. I was thinking you were seeing a card with lots of tuples.

jxjj:
> The cover-fit rendering seems good for images but kind of weird for the placeholder icons (like the blank document icon). Maybe our best bet is to make a set of those icons in 16:9 for the Vue interface?

Is there a way to tell on the vue side when an asset thumbnail is just a placeholder icon (and what file type it is)?

cmcfadden:
Sure, if that'd be easier we could add it to the API. Or we could regex the API if that's just as easy.

cmcfadden:
Erf nevermind, we can't regex the URL.

jxjj:
- [x] Fixed: page jumping when new assets load.
- [x] Added: chip to show the file count for items with more than one file
- [x] Fixed: Search card contents have a max height and contents that exceed the height will be scrollable.
- [x] Added: Show skeleton media card while fetching new assets. Transition fade-in-out when new assets are loaded.

I'll put the icons stuff with a separate PR, since it will sit within the elevator repo.

Let me know how it looks/feels and if there's anything else you notice or that we should address here.

jxjj:
I think we're persisting the scroll position of the content container now so that users can use the back button. But, Im not sure if I like this approach. It seems a little fussy and I worry about missing edge cases. 

An alternative would be to just create a new layout for the search page that doesn't use a container with scroll for the contents.

I might explore that before merging this PR. Thoughts?

jxjj:
Updates: 

- Changed the old default layout to be `<NoScrollLayout>`, and then the new `<DefaultLayout>` has normal scrolling. This let me remove all the extra code for persisting an element's scroll position and instead use vue router's built-in `scrollBehavior()` method.

- Tefactored lazy loading on the search page to use `useScroll` instead of `useInfiniteScroll`. With `useScroll` we're watching `arrivedState.bottom` and loading more results when it's true (and not already fetching).

Thanks again for the feedback.

PR #74, opened 2023-03-28
Add components for rendering static pages like `/page/view/1`

jxjj:
This adds a new `StaticContentPage` component for rendering static page content in Elevator.

Page content is raw HTML, so this includes a `<SanitizeHTML>` component (lifted from another LATIS project) for cleaning up any markup that may be problematic. Under the hood, this component uses the `dompurify` library.

- The page has some very light formatting, which I'm not sold on – so we can (and probably should) tweak this.
- As is typical with user created content, I noticed a lot of inline styles and non-semantic markup in the HTML. There are pros and cons with stripping out these styles. I've chosen to leave them as-is. 

Currently up: <https://dev.elevator.umn.edu/defaultinstance/page/view/3>
Screenshot 2023-03-28 at 6 00 07 PM

Corresponding Elevator PR: https://github.com/UMN-LATIS/elevator/pull/120

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@vueuse/core": "^9.3.0",
     "autolink-js": "^1.0.2",
     "axios": "^0.27.2",
+    "dompurify": "^3.0.1",
     "geojson": "^0.5.0",
     "mapbox-gl-esri-sources": "^0.0.4",
     "maplibre-gl": "^2.4.0",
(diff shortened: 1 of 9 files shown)


cmcfadden (approved):
Looks good - only thing I noticed is that the sidebar nav isn't using router links for pages, but instead is triggering a full page load on each?

jxjj:
Thanks! You're so good at catching those things! I'll take a look.

jxjj:
- [x] Updated the `AppMenu` so that pages use router links rather than normal links.
- [x] Also watching for route changes so that we can close the app menu and recompute current page.

Let me know if you can think of any other page link things that need to be updated!

Issue #73, opened 2023-03-28
Update to Storybook 7 (support for Node LTS v18)


jxjj:
Running the project with node LTS (version 18), fails with this error:

```
➜  elevator-ui git:(develop) ✗ yarn storybook
yarn run v1.22.19
$ start-storybook -p 6006
info @storybook/vue3 v6.5.12
info 
info => Loading presets
ℹ ｢wdm｣: wait until bundle finished: 
node:internal/crypto/hash:71
  this[kHandle] = new _Hash(algorithm, xofLen);
                  ^

Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
    at module.exports (/Users/johnsojr/code/elevator-ui/node_modules/@storybook/vue3/node_modules/@storybook/manager-webpack4/node_modules/webpack/lib/util/createHash.js:135:53)
 ...
```

This is related to a [change with openssl](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported/73027407#73027407) between node 16 and 18. Reportedly, [this issue is resolved in Storybook 7](https://github.com/storybookjs/builder-vite/issues/414#issuecomment-1287536049).

A workaround (other than just using Node 16), is to use `--openssl-legacy-provider` option with node 18.

Like:
```sh 
NODE_OPTIONS=--openssl-legacy-provider yarn storybook
```

jxjj:
Stale, closing.

PR #78, opened 2023-03-31
Add Home Page

jxjj:
This adds support for "Home Page", in the happy case. (Other cases will be handled in separate PR, see: #73 )

With featured asset:
CleanShot 2023-03-31 at 15 32 00@2x

Without featured asset:
CleanShot 2023-03-31 at 15 38 47@2x

Also:
- `<a>` links to `/` in the App Header and App Menu are replaced with router `<Link>`, to avoid full page refresh on navigation.
- Only pages with `includedInNav` prop will be included in App Menu.
CleanShot 2023-03-31 at 15 32 21@2x

- StaticContentPage simplified so that it aligns with the HomePage style
- some theming fixes


cmcfadden (approved):
looks great, nice

PR #81, opened 2023-04-05
Add page to list All Collections

jxjj:
Adds a page which lists all collections. This implements roughly the same UI as the "Classic" interface page:

- the collection list flows top-to-bottom, then left-to-right
- when a collection has children, a chevron will show up. Clicking the chevron will open/close the list of child collections.
- Clicking the collection title, will jump to the search results. (This ultimately will be a collection browse page, but this PR doesn't implement this so you'll see "" as search text.)

Whole page:
CleanShot 2023-04-05 at 11 17 21@2x

With one collection expanded:
CleanShot 2023-04-05 at 11 17 36@2x


cmcfadden:
The expansion effect is a big weird because it grows the boxes for all the collections - any way we can avoid that?

jxjj:
It should grow all the collections on the same row, but not all collections. Are you seeing something different?

Just LMK how you'd like it to look. (I have a note about wanting each collection card within a row to be the same size, but maybe I misunderstood).

jxjj:
Oh, I'm seeing what you're seeing I think. I'll take a look!

jxjj:
Lol, sorry  – I don't know what I was looking at.

Like this, right?

CleanShot 2023-04-05 at 11 38 05@2x

cmcfadden:
yep, that looks better

PR #82, opened 2023-04-05
Add BrowseCollectionPage

jxjj:
CleanShot 2023-04-05 at 18 21 13@2x

This adds a page for browsing a given collection. Example here: <https://dev.elevator.umn.edu/defaultinstance/collections/browseCollection/24>

- Adds a route for `/collections/browseCollection/:collectionId`
- Adds`BrowseCollectionPage`. This page is very similar to the SearchPage, with bulk of the work done by the `<SearchResultsGrid>` component. I first tried a combining both cases in the SearchPage, but it felt too complex, so I separated the pages. 

One note:
I'm a little unclear about how searching works in elevator. Does a `searchId` identify a set of results to a given search at a particular point in time, or does `searchId` identify search parameters and the search itself is run anew each time the page is accessed (so a user always gets the most up-to-date results when they refresh).

Currently, the browseCollection page is set up to do a new search (i.e. get a new `searchId`) when a user refreshes so that they get current search results.  We're also keeping the user on the `/collections/browseCollection/:collectionId` route rather than redirecting to `/search/s/:searchId`. But – but depending on how search works and collections are generally used – this may not be needed or wanted.


cmcfadden:
Good question - searchIds store the search parameters, not the results set. So refreshing a given searchId will get you an up to date set of results. I think we want to redirect them to the /search/s route and handle it that way?

Do you want to incorporate loading the collection description content into this PR, or handle that separately?

jxjj:
Thanks! I'll make the changes. 

And thanks for the reminder about descriptions – totally forgot about them. I'll include in this PR.  LMK if there are any other bits that I might have missed.

jxjj:
Changes:

- collection descriptions added to browsing collections

- `browseCollections/:collectionId` will redirect to the searchId. I explored doing the redirect right in the router, rather than using a basically empty `BrowseCollectionPage` page. However, because we need an  api request to look up the `searchId`  before redirecting, the redirect would be async. Currently, vue-router doesn't support async redirects. So I kept as-is. When redirecting, this uses `router.replace` rather than `router.push` to replace the browse collection route with the searchId route. `router.push` would cause some annoying back button behavior: going back and hitting the previous browseCollectionPage, would immediately redirect back to the searchPage. 

- when browsing a collection, the corresponding searchId is cached. (these should be a 1-to-1 mapping, if I'm understanding the discussion above).

-  `instanceStore` was refactored a little to make it easier to look up a given collection by id. Instead of searching the collection array, and each child collection array, we introduced a computed `collectionIndex` which indexes the collection by it's id and returns a normalized collection.

CleanShot 2023-04-06 at 15 42 11@2x

Latest up on <https://dev.elevator.umn.edu>

cmcfadden (approved):
looks great!

Issue #76, opened 2023-03-30
Search input's clear button sometimes will appear even when the search text is blank


jxjj:

jxjj:
Resolved by #86

PR #86, opened 2023-04-10
fix: persist search input between searches

jxjj:
This resolves:
- #75 
- #76 

These issues were a biproduct of having two different "sources of truth" for the search query: the `searchText` input in the search component and `searchStore.query` in the search store.

This:
- uses `searchStore.query` as the "source of truth" for the search query input and removes the `searchText` ref intermediary
- no longer clears the search query once it's submitted
- switches the search result page header to use `searchEntry.searchText` rather than `searchStore.query` (since searchStoer.query is tied to the input, and not the current results a user may be viewing).
- Shows "All Assets" as a page header if the searchText is empty (and we're not browsing a collection)

Latest up on <http://dev.elevator.umn.edu/defaultinstance>

### DIFF
--- a/src/components/SearchBar/SearchBar.vue
+++ b/src/components/SearchBar/SearchBar.vue
@@ -7,7 +7,7 @@
         label="Search"
         :labelHidden="true"
         placeholder="Search"
-        :value="searchText"
+        :value="searchStore.query"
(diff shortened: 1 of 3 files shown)


cmcfadden (approved):
looks good

PR #87, opened 2023-04-10
Add navigation for prev/next search results within Asset Page

### LINKED ISSUE #56: Asset from Search Results should have Next and Previous navigation
Like this:

Chat with @cmcfadden about details

jxjj:
This adds a navbar with links to the previous and next search result assets on the Asset Page.

Asset page with viewer:
CleanShot 2023-04-10 at 17 14 33@2x

Meta data only page:
CleanShot 2023-04-10 at 17 15 20@2x

This version uses the searchStore for determining the next/prev search results, rather than doing another API query. When the user is within LOAD_MORE_THRESHOLD (currently 5) results of the last loaded result, more will be fetched.

currently up: <https://dev.elevator.umn.edu/defaultinstance/>

Closes #56

### DIFF
--- a/src/components/Button/Button.vue
+++ b/src/components/Button/Button.vue
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="componentType"
-    class="button inline-flex items-center gap-1 no-underline hover:no-underline rounded justify-center leading-none transition-colors ease-in-out group"
+    class="button inline-flex items-center gap-1 no-underline hover:no-underline rounded justify-center leading-none transition-colors ease-in-out group cursor-pointer"
     :class="{
       'button--primary px-4 py-3': variant === 'primary',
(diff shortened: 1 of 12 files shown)


cmcfadden (approved):
looks great! Do you think it would make sense to have the number of results (1 of XXX) be a link back to the search results?

jxjj:
Love that idea!

jxjj:
Two updates:
- the current result index (e.g. "101 of 253") will link back the the search results
- the search results page will scroll to the indexed object so that it's in view (e.g. it will scroll to object 101), rather than just returning to the top of the page.
- If the object isn't on the page for some reason (maybe a refresh, where and the result set no longer contains 101), it will just scroll to the top.

Let me know how that looks and feels to you. I _think_ it's better to jump to the object, but it might feel less surprising to users to just return to the top?

Issue #58, opened 2022-12-12
Add different views of search results, e.g. map, timeline


jxjj:
- [ ] Grid
- [ ] List
- [ ] Map
- [ ] Timeline
- [ ] Gallery

![Screenshot 2022-12-12 at 10 38 21@2x](https://user-images.githubusercontent.com/980170/207102307-554f78c1-7407-4497-8577-f394dde09a4f.png)

PR #106, opened 2023-04-12
Add tabs and List View to Search Results page

jxjj:
Part of #58.

This:
- adds tabs for switching between different search results views
- adds a list view to the search results page
- adds a Back to Top button
- disables dark and hotdog themes (for now). Different themes with different typography cause the app-header and sticky tabs to have different heights, which makes layout a bit fussy. I think I'd prefer to circle back than troubleshoot now.

Default (Grid) view with tabs:
CleanShot 2023-04-12 at 17 20 17@2x

When a user scrolls, the tabs will `stick` to the top:
CleanShot 2023-04-12 at 17 20 41@2x
d37b80b-6f16-450b-9d57-7726ab1f25b4.png">

List view:
CleanShot 2023-04-12 at 17 21 11@2x

This doesn't have any view state persistence yet. That is, the back button will revert to the default grid view. I could add that here or in a separate PR (this PR was already getting big).

Up here for testing: <https://dev.elevator.umn.edu/defaultinstance/>


cmcfadden (approved):
Looks good. In the current site we use an anchor to determine the tab, so that folks can link into a specific view. (IE, adding #grid, #list) ... that might be preferable vs trying to specifically track state around the tabs

jxjj:
Thanks and agreed on tracking state around tabs. We're currently using the anchor  when linking back to a particular object, like:
<https://dev.elevator.umn.edu/defaultinstance/search/s/835947d0-f709-4945-b79d-28fc189b5979#object-575ee42bba98a87472f3d4db>

So, maybe a query param for the view? 
Or perhaps it makes more sense to move the object to the query param, and do the results view as a hash?

A: `?objectId=12345#list`
B: `?view=list#object-12345`

I think I might like `A` better? What do you think?

cmcfadden:
Yeah I think A

PR #110, opened 2023-04-17
Search Results Timeline View

jxjj:
ScreenShot 2023-04-17 at 17 44 47@2x

This adds a timeline view of search results using TimelineJS within the `<SearchResultsTimeline>` component.

Example up here: <https://dev.elevator.umn.edu/defaultinstance/search/s/d3f91f49-cd8b-451e-b794-c0a05e4852fa?resultsView=timeline>

The bulk of the work is in the conversion from search results to TimelineJS slides. The existing `prepTimeline` function instructive, but my initial attempts to migrate it directly wasn't productive. (The current version had dom manipulation, global variables, and other things that caused linter warnings and type errors.)

So, when developing this version, I added`jest` for testing, then using the Obama Collection of Speeches for input and the slides results returned from the dev server for the expected output (held in the `compiledDate.events` variable) I created some helper functions to generate the results. (I needed to modify some of the expected output to fix the date mismatches. See below).

I'm almost positive that there's functionality in `prepTimeline` that's not incorporated into this PR. I'd love to walk through it and clarify the intent – maybe bringing those cases over in another PR?

## Other Note:
- I noticed on dev server using the legacy theme, there were cases where dates were mismatched in the slide results. For example, Obama's Counterterrorism speech had a `display_date` of  `05/23/2013`, but the `day` prop was 5. Not sure if it was a bug or a typo or something:

- `less` is added as dependency for TimelineJS
- added some typings were added for TimelineJS objects.


cmcfadden (approved):
I think this seems right and is fine to merge. I think the biggest functionality you're missing is the separate handling for Really Old Stuff (geo time) where we drop the idea of months/days rendering things with a different scale? I'm not sure any site has actually used that though, so it's really not important.

jxjj:
Nice, thanks!

PR #112, opened 2023-04-21
Add Map view of search results with data point clustering

jxjj:
ScreenShot 2023-04-21 at 15 30 23@2x

<https://dev.elevator.umn.edu/defaultinstance/search/s/d34a26d0-0239-4f63-8bce-3a489844bd18?resultsView=map>

This adds a map view to the search results.

### Clustering

The bulk of this PR is reworking `<Map>`, `<MapMarker>` and `<MapPopup>` to support clustering. 
Unfortunately, there's no easy way to just turn clustering on using the Marker API. Instead, to draw clusters with MapLibreGL, we need to:
1. Add a data source of markers as a GeoJSON FeatureCollection to the map
2. Add 3 layers to the map: clustered markers, text labels for the number of points within a cluster, and normal points not within a cluster.

At first, I thought I could just get a list of markers added to the map instance, and massage the data into the a datasource and layers. But, it appears [there's no easy way to get a list of a markers directly from the map instance].

So, this PR inverts the control of these components. That is, instead of the `<Marker>` component making a Marker instance and adding itself to a map instance, the `<MapMarker>` would tell it's parent `<Map>` it needs to add a marker using a Map's `addMarker` function. (And same with a `<MapPopup>`: it invokes `addPopup`, from it's parent `<MapMarker>`.

All this is using vue's `provides`/`inject` like before, but instead of passing down the map instance as `MapContext`, we're passing down functions that the children can invoke. 

As a result of all this inversion, the `<Map>` component holds a lot more state: the map, any map markers, and popup containers, and a bit more logic to handle the marker and popup rendering. Which is to say, t`<Map>`'s `onMounted` method is pretty big. I considered refactoring, but I worry that I'd be introducing abstraction and indirection, without improving readability. 

### Changing Map Styles with Layers and Data Sources

One notable thing about using layers and data sources: when changing map styles (e.g. from Street to Satellite), we may need to re-add the data source and layers to the map. So, it's not good enough to just add Data Source and Layers on `load`. We changed to adding layers and styles on the `styledata` event, which is emitted with each style change.

### Fonts and Rendering Text Layers

After inverting the control of the markers to `<Map>` things should have been similar to the [MapLibre  clustering example](https://maplibre.org/maplibre-gl-js-docs/example/cluster/). It was except for one pitfall I'm noting here (hoping I don't forget in the future).

The example's text layer example seemed to cause nothing to render other than the blank map. Removing the text layer made clustering work fine. It seems that the issue is related to different fonts or styles available to ArcGIS. Looking at [this ArcGIS example using Clustering with Mapbox], I noticed they were using a different font: `Arial Bold`. Changing to `Arial Bold` suddenly made everything render.

(In my debugging, I thought fonts might be an issue, so I tried `Arial` thinking this would be the most vanilla of fonts. Arial doesn't work though, only Arial Bold.)

### Final Notes

- This PR contains a few other helpers for things like calculating bounding box for the map. Most of this is lifted directly from Camino.
- With all of this inversion of control, I'm a little more concerns about memory leaks. I think vue and MapLibre will handle most cleanup. I've tried to do additional clean-up too, but I may have missed something.
- There are places where Typescript will complain about a MapLibre Map and Marker instances being "deeply nested and possibly infinite" when assigning the instance to a ref. I don't recall running into this issue with MapboxGL – I suspect there's something off with the MapLibre typings? Or the play between Ref type and the MapLibre type? I'm not sure. Rather than chasing this down, I used typecasting (`as MapLibre`) where I could – or failing that, I just deactivated typescript for the line using `// @ts-ignore`.


cmcfadden:
Could we add a bit more height on the map? I'm also seeing something a little weird- this is after loading all results on the Obama collection. Not sure if it's a data problem, or if we need to be fuzzing the coordinates a bit (I think the only one did that?). 

Screenshot 2023-04-24 at 9 03 44 AM
Screenshot 2023-04-24 at 9 03 46 AM

cmcfadden (approved):

jxjj:
Note to self: merge, don't rebase after a PR. 

Latest on <https://dev.elevator.umn.edu/defaultinstance/search/s/681f8a51-02bf-481b-85a3-f227c981dbab?resultsView=map>

## Update

I tried adding spidering with [MapboxGL Spiderifier](https://github.com/bewithjonam/mapboxgl-spiderifier), but just couldn't get it to work... so I just went the offset route: when a point is added to the map, there's a check to see if it's close to other points. If it is, it's given an offset of 0.0001 * number of super close points added.

For example, here are two Obama speeches in the same location:
ScreenShot 2023-04-25 at 21 57 24@2x

I opted for a deterministic offset rather than random, which might not be the right approach... but seemed to work well  for our test data set. I'm calling it "good enough," but I'm happy to change if you think this will be an issue for some of our users.

Also:
- removed the Results count below the map (similar to timeline page)
- fixed the bottom results count – it was still using "Load More" instead of "Load All".

PR #115, opened 2023-04-26
Fix App Menu `Edit` doesn't appear

jxjj:
When there's no asset loaded in the assetStore, the Edit menu won't appear in the App Menu since this `assetId` is null.

This removes the check for a truthy `assetId` from `<AppMenu>` and instead checks within the `EditNavSection` so that users still have an option to Add Assets or view all their assets if they `canManageAssets`.

### DIFF
--- a/src/components/AppMenu/AppMenu.vue
+++ b/src/components/AppMenu/AppMenu.vue
@@ -14,7 +14,7 @@
       </AppMenuItem>
 
       <EditNavSection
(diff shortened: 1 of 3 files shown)


jxjj:
Hmm... I probably should also be clearing the activeAssetId when navigating away from the Asset View page, so that user's can't edit/delete an asset that they're not looking at.

e.g. Home Page -> view asset -> go to Home Page (asset is still marked as active in the store).

Issue #101, opened 2023-04-11
Theme switcher, move to app menu


jxjj:

jxjj:
Resolved in https://github.com/UMN-LATIS/elevator-ui/issues/101

Issue #85, opened 2023-04-05
Nit: User initial box looks weird


jxjj:
It's sometimes a rectangle. Nobody likes rectangles. Should probably be either a rounded square or circle. Maybe only one initial? And is color needed? Now the the design system is settling, maybe simplify? Chat with @cmcfadden.

<img width="64" alt="CleanShot 2023-04-05 at 18 48 48@2x" src="https://user-images.githubusercontent.com/980170/230238244-7765059f-e089-4c0c-a47c-c41cc4aef68f.png">

jxjj:
Resolved in #116

PR #116, opened 2023-04-26
App Header: Move Theme Selector, Fix Avatar Proportions (stacked on #115)

jxjj:
Resolves:
- #101 
- #85 

This updates the app header:
- moves the Theme selector to App Menu flyout.
- Fixes the Avatar initial proportions and simplifies the look: circular, without background color, drops the chevron, adds "Signed in as..." to the dropdown menu. Also changes the first initial font to be generic `sans-serif`. Some fonts (like open sans) have uppercase glyphs (like "J") that descend below the baseline and look off-center.
- Since the Avatar dropdown now shows "Signed in as", this info in dropped from the flyout App Menu.
- Changes the Search Input colors from neutrals to transparent blacks. The grayscale text looks off when the background has a color, like in the Folwell theme. Neutral black mixes better with the bg color.

Stacked on #115 

Live on dev: <https://dev.elevator.umn.edu/defaultinstance/>

## Screenshots

Not Logged in:
ScreenShot 2023-04-26 at 14 46 16@2x

Logged in. With updated avatar, and "Logged in as..." in dropdown:
ScreenShot 2023-04-26 at 14 45 41@2x

Updated app menu. Theme Selector across from "default instance":
ScreenShot 2023-04-26 at 14 45 53@2x

### DIFF
--- a/src/components/AppHeader/AppHeader.vue
+++ b/src/components/AppHeader/AppHeader.vue
@@ -10,13 +10,7 @@
       class="flex-1 w-full max-w-lg"
     />
     <div class="flex gap-2 items-center">
-      <ThemeSelector
-        v-show="config.instance.theming.enabled"
-        class="hidden sm:block"
-        :defaultTheme="config.instance.theming.defaultTheme"
-      />
       <AuthDropDown
-        class="hidden sm:block mr-2"
         :currentUser="currentUser"
         :instance="instanceStore.instance"
       />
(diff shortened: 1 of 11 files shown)


cmcfadden:
Seems like edit->add asset is broken?

jxjj:
What are you seeing?

We haven't built any Add Asset stuff yet – we're just redirecting to the modal page (which is admitted broken looking), and then the user gets redirected to the old website. 

Or are you not even getting this redirect?

cmcfadden:
ah my bad - I totally thought we'd had an actual modal of the add asset stuff. Sorry

cmcfadden (approved):

jxjj:
No worries. I'll add a GH issue so that it's on the list.

PR #118, opened 2023-05-02
Add gallery view for search results

### LINKED ISSUE #58: Add different views of search results, e.g. map, timeline

jxjj:
This adds a gallery view of search results:
- The gallery creates slides for search matches, any files within the asset (that is, any objects with a defined `fileId`), and any related assets items found in the asset's `relatedAssetCache`. 
- While adjacent results are loading asynchronously, placeholder slides will be used.  
- When the `activeSlideIndex` is within 10 of the last slide, the gallery will load more search results.

(At least) one bug that needs to be resolved: with some searches on dev, like "Test", the gallery with throw an error and the active slide index won't update properly in the main slider. I'm probably making an incorrect assumption about the shape or existence of some match result or asset data.

Currently up on dev for testing: 
https://dev.elevator.umn.edu/defaultinstance/search/s/7eeb0e19-1f48-4806-a22c-238c2aa39199?resultsView=gallery

image

Closes #58

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -113,7 +113,9 @@ const router = createRouter({
       name: "error",
       path: "/error/:errorCode",
       component: () => import("@/pages/ErrorPage/ErrorPage.vue"),
-      props: true,
+      props: (route) => ({
+        errorCode: parseIntFromParam(route.params.errorCode),
+      }),
     },
     {
       path: "/:pathMatch(.*)",
--- a/src/router.ts
+++ b/src/router.ts
@@ -31,6 +31,10 @@ const routesForTesting: RouteRecordRaw[] = [
       objectId: route.hash?.substring(1),
     }),
   },
+  {
+    path: "/test/slides",
+    component: () => import("@/pages/TestPages/SlidesTestPage.vue"),
+  },
 ];
 
 const router = createRouter({
--- a/src/router.ts
+++ b/src/router.ts
@@ -31,10 +31,6 @@ const routesForTesting: RouteRecordRaw[] = [
       objectId: route.hash?.substring(1),
     }),
   },
-  {
-    path: "/test/slides",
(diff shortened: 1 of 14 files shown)


src/router.ts
```diff
@@ -31,6 +31,10 @@ const routesForTesting: RouteRecordRaw[] = [
       objectId: route.hash?.substring(1),
     }),
   },
+  {
+    path: "/test/slides",
```
cmcfadden, line 35: do we want to keep this route?

cmcfadden:
Is a fullscreen mode still on your todo for this, or am I missing it?

One issue I'm seeing is that if you grab the slider at the bottom and move, it indicates it's changing the active item, but doesn't actually? Is that an easy-ish fix?

May-02-2023 13-34-21

jxjj:
Thanks for the feedback and testing. I think this is ready for another round of review:

Changes:
- Fixed issue where gallery would sometimes throw. Turns out I made one too many placeholder slides.
- Added fullscreen toggle.
- Shifted layout a bit to accommodate fullscreen. It was getting busy at the top, and the prev/next there were redundant with the prev/next by the thumbs at the bottom, so I dropped the top buttons.
- Now sliding thumbs will update the main slide properly. One caveat: we're waiting for the drag to complete before updating the main slider to avoid loading extra object viewers.
- Added keyboard navigation with left/right arrows.

image

src/router.ts
jxjj: dropped

jxjj:
Closes:
- #58

cmcfadden:
looks great!

cmcfadden (approved):

PR #128, opened 2023-05-05
add sort select to results display

jxjj:
This adds sorting for search results. Sort is added as a query param so that a link can be shared and users will see the same sorted results.

ScreenShot 2023-05-05 at 09 38 04

There's a little churn on the search components – especially the gallery component – as I was playing with formatting options to accommodate the results count, the sort select, and the fullscreen button. Ultimately I landed on keeping sort and results count together and putting them in a very slightly gray box so that the extra whitespace doesn't look quite so off.

<https://dev.elevator.umn.edu/defaultinstance/search/s/19413d13-aaf2-4067-8290-114b38787ee3?resultsView=grid&sort=title.raw>

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -251,11 +251,6 @@ const api = {
 
     // this param gets searchID without all the results
     params.append("storeOnly", "true");
-
-    if (opts.sort) {
-      params.append("sort", opts.sort);
-    }
-
     const res = await axios.post<SearchResultsResponse>(
       `${BASE_URL}/search/searchResults`,
       params
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -10,6 +10,7 @@ import {
   ApiInstanceNavResponse,
   ApiStaticPageResponse,
   FileDownloadNormalized,
+  SearchSortOptions,
 } from "@/types";
 import { FileMetaData } from "@/types/FileMetaDataTypes";
 import { FileDownloadResponse } from "@/types/FileDownloadTypes";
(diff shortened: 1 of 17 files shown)


cmcfadden:
This is looking good. There's one special case I neglected to describe.

If the search text is blank, the results are going to be displayed in alphabetical order rather than best match, since best match with an empty search is not defined.

jxjj:
Updates:
- When the query text is blank (e.g. when browsing collections), the default sort will be "Default Title" instead of "Best Match". "Best Match" will be also be a disabled option (I'm not sure if we want this or not, lmk)
- I noticed that collection info wasn't persisting on sort change, so this fixes the persistence issue.

## Notes on flashing browsing collections title

When changing the sort while browsing collections, the title and description will "flash" (disappear and reappear) until the new sort results are returned.  This is because the collection title is derived from `searchStore.searchEntry`, which is reset with each new search. It's not awful. Just annoying.

I tried keeping the old search entry until the new one arrived, but that led to problems when browsing collections – a user would see the wrong collection name until the search completes. I think I probably need to rethink `searchStore`, and move away from computing info from search entry, maybe keeping more state for searches in flight?

For now, I'm just showing a "skeleton" while the waiting for the search entry to reduce the amount the page results jump around. 

I suspect there might be other bugs related to searching/sorting/browsing collections. My current thinking is that these might be better addressed when implementing advanced search, since we'll be doing collection search work as part of that.

cmcfadden (approved):
looks great

PR #140, opened 2023-05-24
Advanced Search: filter by text fields

jxjj:
This update introduces filtering by specific text fields to advanced search. Users can now add multiple text field filters, with the choice to toggle between AND/OR boolean operators. 

image

ScreenShot 2023-05-23 at 21 03 46

While this PR covers only text field types, future PRs will expand to other types. The choice of text fields as a starting point is due to their simplicity and fewer API calls for displaying user options.

Potential search/sort fields are kept in `instanceStore`, while  `searchStore` keeps the field filters and handles creation, updates, and deletions. 

Search filters are assigned unique UUIDs to accommodate multiple filters for the same field. They are stored in a Map for faster value updates and are translated to a format suitable for the search POST request via `searchStore.searchRequestOptions`

In the UI, boolean operators OR/AND are displayed only when more than one field filter is present. They are interactive, enabling the user to toggle between OR/AND by clicking on them.

I'm seeing some unexpected results in some searches. For example, when searching collection "Geotime Test" for title "Dino", I don't get any results. Same with "Obama Speeches" and "Afghanistan". I'm not sure why.

Next up: fuzzy search and extend filtering capabilities to select and other field types.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -236,6 +236,7 @@ const api = {
     const res = await axios.get<ApiInstanceNavResponse>(
       `${BASE_URL}/home/getInstanceNav`
     );
+
     return res.data;
   },
 
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -10,8 +10,8 @@ import {
   ApiInstanceNavResponse,
   ApiStaticPageResponse,
   FileDownloadNormalized,
-  SearchSortOptions,
   LocalLoginResponse,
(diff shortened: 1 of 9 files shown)


cmcfadden:
The "Dino" assets in geotime test don't have a title field, they have a "speech title" field, that's why those aren't working the way you expect

cmcfadden (approved):

PR #145, opened 2023-05-26
add support for instance-level configuration of themes

jxjj:
This adds front-end support for instance configuration of theme settings. The app will use theme settings on `window.Elevator.config.instance.theming`:

- `theming.enabled` – enables or disables theming of this instance.
- `theming.availableThemes` – list of supported themes for this instance like `['folwell', 'light', 'dark']`. A supported theme must have a corresponding CSS file (e.g. `folwell.css`).
- `theming.defaultTheme` - The default theme used on this instance if none is set or if the user's set theme is unavailable (e.g. if the available themes changed)

The localstorage theme key is `theme-{{BASE_URL}}`, so multiple instances on the same domain can have separate keys.

Other changes:
- The ThemeSelector component we refactored, removing the `useTheme.ts` composable. (This was introduced so that we could import types, because you couldn't import types from a vue file, but you could import from a ts file. This was recently changed, so isn't necessary. Plus, we simplified theme type to just be plain string, instead of a object).
- [`deepmerge` dependency](https://www.npmjs.com/package/deepmerge) was introduced so that we could deep merge objects with arrays. It's tiny and does the trick.

The `vueTemplate` on the backend just has settings hard coded in the `window.Elevator.config` object for now.

Available for testing: <https://dev.elevator.umn.edu/defaultinstance/>


cmcfadden (approved):
this looks good. Let's not merge it until I finish the backend work, just in case we unearth any changes needed there?

PR #147, opened 2023-06-01
Add select field type to advanced search of fields (stacked on #146)

jxjj:
This adds support for advanced searching using select fields. This is stacked on PR #146.

Here I am randomly clicking select stuff:

A few related changes:

- There was an issue where the vue app was double mounting and causing all sorts of mysterious issues with the search store, (only the cryptic error ` Uncaught (in promise) TypeError: Cannot read properties of null (reading 'isCE')`. The cause seemed to have something to do with loading dynamic components in vue router or how vite (or rollup) is bundling them. Maybe related to the versioned `script.js?version`? I'm not sure. But, I've switched back to regular imports in vue router for now, thinking we can tackle it when we finish up the versioning stuff on the elevator-ui side.

- Sometimes loading or refreshing a search page would error. This was a race condition between the `searchStore` and the `instanceStore`. When the searchStore received search results containing field information before the instance store resolved (which is our source of truth for field labels, field types, etc), the searchStore would throw an error. In this PR, I'm explicitily loading instanceStore in the `App` component and only mounting the App contents when `searchStore.isReady`.

- The `api/index.js` was getting large, so I split it in two. `api/fetchers.js` is responsible for all the axios/fetching stuff. `api/index.js` handles caching the results and massaging data into whatever format is needed by the frontend components if needed.

One question: for some select fields, the response from `search/getFieldInfo` included an empty string in the array. For example, here's the response for the System select field on dev:
ScreenShot 2023-06-01 at 14 05 19@2x

I assumed that this was an oddity of data on dev, so I'm just ignoring empty strings as select choices. But, let me know if this is a special case we need to handle.

Available on <http://dev.elevator.umn.edu/defaultinstance> for testing.


cmcfadden:
The empty strings in those drop-down are intentional and user-created. That's how you tell the system there's no value / no selection for that field for a given asset. So it's essentially how you enter "null". Users can omit that empty string if they want to force a selection.

jxjj:
Readability. The file was big and this seemed like a sensible seam

cmcfadden:
got it - just making sure I wasn't missing something

cmcfadden (approved):

jxjj:
Update:
- Added the empty string option back in. The first option is selected by default if no value is set.

Questions:
1. As a user, I'm not sure if I would know how to interpret a blank select. (or even that it was a legit option). Do you think we should label it differently than just a blank? Like a null or dash? Or is it fine as-is?

2. Also, I think I might be misinterpreting what the search field filters do (or what I've been calling "field filters")? I was thinking that they filter the results of the text search. But, if that was the case, we should always have fewer results than the plain text search.

A plain "test" search returns 238 results:
ScreenShot 2023-06-01 at 16 09 44@2x

But a "test" search with a filter when the "System" field  equals "NTSC" returns 6784 results:
ScreenShot 2023-06-01 at 16 09 21@2x

What am I misunderstanding? Should we be calling this something other than filters?

cmcfadden:
I'll spend some time reviewing the search logic tomorrow. I think that it does an OR rather than an AND between the field specific search and the general search field. I agree that the blank isn't clear - I'm also not sure it's actually implemented properly on the backend, so I'll look into that.

jxjj:
Thanks for looking into it! Not a blocker or anything. I'll change the blank to a `-` for now, and merge. If the boolean is more global than the fields section, we can shift around some bits in the design.

PR #148, opened 2023-06-02
Change build to generate a `manifest.json` so that it can use hashes in file name

jxjj:
This resolves an issue with asset caching beyond the main js and css files using the approach outlined in the [Vite Backend Integration Guide](https://vitejs.dev/guide/backend-integration.html).

This changes vite's build step to generate a `manifest.json` file can be used by the backend for cache-busting.


cmcfadden:
ah yeah, makes sense :)

cmcfadden (approved):

PR #152, opened 2023-06-05
load theme on app mount and visual fixes for advanced search

jxjj:
- sets initial theme on app mount, which fixes an issue where the theme wouldn't be set until until the app menu is opened
- reduce the visual hierarchy of dropdown "Add" menus in the advanced search
- Fix column height on home page so that it stretches the full page. (A regression after adding the skipnav link and wrapping the main content in a <main> element).


cmcfadden (approved):
looks good!

PR #155, opened 2023-06-06
Add cascading multiselect to advanced search

jxjj:
Adds a cascade select component and multiselect field options to advanced search. Also fixes a bug where the search boolean wasn't set when we received search results.

ScreenShot 2023-06-06 at 13 05 11


cmcfadden (commented):
Looks good. Can you add an alphabetical sort to the values? The current version applies that so we should probably maintain it.

jxjj:
Updates:
- options are now sorted
- fixed an issue where switching field from one multiselect to another would keep the same options.

PR #156, opened 2023-06-09
Add `Any Date` to Advanced Search

jxjj:
This lets a user search by `Any Date` in the advanced search.

### Background:
- Any Date is a special global search field, not a specific field like other fields like "Title" or "Author". It doesn't have a real `fieldId`, it isn't returned with other fields with `getInstanceNav`, and it's values shouldn't be included with other specific fields in the search query.  
- We want it to appear like it's just another specific field filter to the user in the UI. That is, we want the user to be able to select it from the dropdown with other fields, and update it's values like other fields.
- The solution should accommodate other general fields. There's an Any Location filter too, which will need to use a similar pattern.

### The Approach

We keep the data for the any date filter in the searchStore in `filterBy.globalDateRange` and not mixed it in with `specificFieldsMap` to make it a bit easier to differentiate from real specific fields.

When generating the list of filter rows in the UI, it's handy to have the date range filter data in the same shape as the other specific filters. So, the `searchStore.globalDateRangeAsFilter` getter will take care of this, assigning a special id and field id `GLOBAL_FIELD_IDS.DATE_RANGE`. 

Once the specific filters and the global Any Date filter are the same shape, we can merge the filters together into one list of rows and then [loop through that list](https://github.com/UMN-LATIS/elevator-ui/compare/feature/adv-search-any-date?expand=1#diff-e22e9d1962eaafa9255bf430af5673d605a11042465846ae73aeef9a80a790e6R20-R26). Specific field filters are still handled by the `<FilterByFieldsRow>`.  Meanwhile the global Any Date filter is handled by the `<FilterByGlobalDateRow>` component. 

The components are very similar, but have different actions on update and delete (and don't allow field changes).

We need some way to make sure the merge specific and global filters are listed in the order the user added.  (Previously we didn't have to do anything, since the specific fields were coming from the same Map interable).  So, we've added a new `createdAt` property to each created filter, and then sort the filters by their `createdAt` date.

#### InputGroup `v-model`

It's handy to just use `v-model` on the Any Date filter's start and end date, so `<InputGroup>` was updated to use `modelValue` instead `value`.

#### Dates as Unix timestamps

The `parseDateString` function was ported over from the legacy UI and some tests were added to check the math matched the datetimes that the legacy ui was generating (especially dates pre-1970).

#### Filter counts and clearing

filter count and clear functions were updated so that the globalDateRange was included. (globalLocation was also added, but globalLocation UI stuff isn't included with this PR).

#### Filter UI Layout changes

With two different row components with different field types (select, input, cascade select), there was a lot of opportunity for misaligned UI elements.

I moved from `flexbox` to `grid`. And fixed some issues with smaller screensizes:

Advanced search at multiple screen width:

ScreenShot 2023-06-09 at 10 05 16@2x

ScreenShot 2023-06-09 at 10 03 36@2x

ScreenShot 2023-06-09 at 10 03 56@2x

#### Scrolling with a large number of fields

I also updated the advanced search modal so that the filter by section will scroll when it gets too big.

#### Dropdown menu positioning with FloatingUI

CSS positioning of a dropdown within a scrollable positioned element gets tricky. We want the dropdown to be able to escape the container when opened, not just add more height to the scroll:

ScreenShot 2023-06-09 at 10 10 17@2x

Also we want the dropdown to flip if needed:
ScreenShot 2023-06-09 at 10 14 23@2x

This PR also pulls in the HeadlessUI Float package for smart dropdowns.

Why this package? 

Bootstrap uses PopperJS for handling their dropdown menu. Popper's latest version was rebranded as [Floating UI](https://floating-ui.com/) (I love their webpage) and has [integrations](https://github.com/ycs77/headlessui-float) with [HeadlessUI](https://headlessui.com/) (which we're using to create the accessible dropdown).

The `<Float>` element makes the work of smart dropdowns almost trivial.

On <https://dev.elevator.umn.edu/defaultinstance/> for testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "@fontsource/material-icons": "^5.0.1",
     "@fontsource/open-sans": "^5.0.1",
     "@fontsource/work-sans": "^5.0.1",
+    "@headlessui-float/vue": "^0.11.2",
     "@headlessui/vue": "^1.7.13",
     "@knight-lab/timelinejs": "^3.9.2",
     "@mdx-js/react": "^2.3.0",
--- a/src/api/fetchers.ts
+++ b/src/api/fetchers.ts
@@ -16,10 +16,8 @@ import type {
   ApiInstanceNavResponse,
   SearchRequestOptions,
   LocalLoginResponse,
-  SearchableFieldFilter,
   ApiGetFieldInfoResponse,
-  SearchableSelectField,
-  SearchableField,
+  SearchableSpecificField,
 } from "@/types";
 import { FileMetaData } from "@/types/FileMetaDataTypes";
 import { FileDownloadResponse } from "@/types/FileDownloadTypes";
@@ -243,7 +241,7 @@ export async function loginAsGuest({
 }
 
 export async function fetchSearchableFieldInfo<T = ApiGetFieldInfoResponse>(
-  field: SearchableField
+  field: SearchableSpecificField
 ) {
   const formdata = new FormData();
   formdata.append("fieldTitle", field.id);
--- a/src/api/fetchers.ts
+++ b/src/api/fetchers.ts
@@ -158,24 +158,23 @@ export async function fetchSearchId(
   query: string,
   opts: Omit<SearchRequestOptions, "searchText"> = {}
 ): Promise<string> {
-  const params = new URLSearchParams();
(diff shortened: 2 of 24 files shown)


cmcfadden (approved):
looks good

PR #159, opened 2023-06-09
Clear search filters on static pages

### LINKED ISSUE #157: Advanced filters don't repopulate without a refresh
1. Do a search using advanced filters
2. click on an asset
3. click the back button
Result: advanced search filter isn't drawn in the search box

Now, refresh page: filter reappears

Same thing happens when clicking the "X of X" in the asset.

https://github.com/UMN-LATIS/elevator-ui/assets/211655/8a36d048-70cb-41f3-a102-38261a7dc595

jxjj:
This fixes an issue where search filters would clear when navigating back to the search results page from an asset page.

1. Instead of always clearing the search filters when leaving the search results page, the filters when be cleared whenever arriving on a page other than the search results page or asset view page. (Some route names are added to `router.ts` to make route checking more readable.)

2. We always do a `searchStore.search` when arriving on a page. 

Previously, we would do a check if the search was already in flight, and then skip running a `searchStore.search` to prevent double search requests. The double request was caused by the search submit handler kicking off a search and then redirecting to the results page. But also arriving on the results page would also kick off a search, unless we checked if a search was already in flight.

To allow us to always do a search while also preventing double-requests, this splits up the search responsibilities into two functions:
- `getSearchId` for getting the search ID ,
- `search(searchId)` for getting results given an id. 

The search form submit handler was changed to only get the searchId and redirect to the results page. And the results page will now always get the results given the search id.

I've left in place caching `search` results for a given ID at the API layer.

Resolves #157 

Also:
- fixed a typo `timline` instead of `timeline`
- removed some unused imports and commented out code blocks.

### DIFF
--- a/src/components/SearchBar/SearchBar.vue
+++ b/src/components/SearchBar/SearchBar.vue
@@ -34,14 +34,13 @@
   </div>
 </template>
 <script setup lang="ts">
-import { onMounted, ref, watch } from "vue";
+import { ref } from "vue";
 import InputGroup from "@/components/InputGroup/InputGroup.vue";
 import AdvancedSearchForm from "@/components/AdvancedSearchForm/AdvancedSearchForm.vue";
 import { useSearchStore } from "@/stores/searchStore";
 import { useRouter } from "vue-router";
 import SearchTextInputGroup from "./SearchTextInputGroup.vue";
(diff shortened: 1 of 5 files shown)


cmcfadden (approved):
👍

PR #161, opened 2023-06-13
Advanced Search: Search by location

jxjj:
This implements location search in advanced search.

ScreenShot 2023-06-12 at 19 01 50@2x

Circle drawing calculations by [@turf/circle](https://www.npmjs.com/package/@turf/circle), which is pretty rad. It even handles skewing as the center gets closer to the poles.


cmcfadden (commented):
I like the look, but I'm not sure the slider/circle works so great from a usability perspective. It makes it really hard to do smaller radius searches - 2 miles, or 0.1 mile or something. Maybe just move that to a text entry? I don't think there's going to be a nice scroll-bar approach that can encapsulate both "show me items on the same block as this one" and "show me items in the same country as this one".

jxjj:
Updates:
- Using text input instead of range.

cmcfadden (approved):
I can input values using LaTeX right?

PR #176, opened 2023-06-16
Improve Add Asset UI

jxjj:
This adds a dropdown menu for users to add an asset with a given template.

Add Asset Flow

Originally, I tried using a modal like the legacy site, but this started adding complexity. Since we're focus trapping the app menu for accessibility reasons, if we open a modal, we'd need to change the focus trap to trap focus on the modal and not trap focus on the app menu. This seemed to require state tracking of both the app menu open/close state and the modal open/close state. We would also likely need to do some work with event handlers listening for escape keys. 

Instead, I chose to make `Add Asset` a dropdown within the App Menu. When a user chooses the asset's template, a hidden form will be created and submitted using the `submitHiddenForm` helper, which redirects to the legacy add asset page.

One additional change: I renamed the `Edit` submenu  to `My Assets`, which I think is a bit clearer.  "Edit" seemed confusing in the context of the home page – it made it seem like it was a menu to edit the page, and I wouldn't find asset management functionality inside. "Assets" sounded a little too general. LMK if there's a better name, or you want me to change it back.

On <https://dev.elevator.umn.edu/defaultinstance>


cmcfadden:
I feel like "my assets" becomes confusing once you're actually on an asset though - if I want to edit the current asset, I may or may not have been the one who created it. So, looking for the edit menu under "my assets" doesn't seem right.

Manage Assets would be reasonable. Otherwise, I'd just stick with the existing "edit". 

The add asset interaction seems a little weird, since the chevron goes down but the menu pops up. Ditch the chevron or change to another icon? 

We can see how that interaction goes and see if users feel like they're missing a secondary confirmation. I know when I'm working on sites with a large set of templates, I sometimes pick one from the list then pause for a bit before I hit the submit button.

jxjj:
updates:
- changed from dropdown to interstitial page at `asset/create`
- changed "My Assets" to "Manage Assets"

cmcfadden (approved):

Issue #105, opened 2023-04-11
Add generic error modal when there's some unforeseen issue that we're not catch


jxjj:
Model after ChimeIn

Issue #84, opened 2023-04-05
Surface errors better


jxjj:
Something better than `console.error`? 
Maybe toast style notifications when there's an error. (This might be ready to go with the `notificationStore`, it just needs to be used?

PR #179, opened 2023-06-23
Add better error handling

jxjj:
This improves error handling for API errors that could be encountered.

An global error modal is added at the top app level. This modal will appear whenever an error is set in errorStore. Axios is given an interceptor, which will set the error if some api error is encountered (e.g. the server returns an unauthorized or forbidden or not found response).

Resolves:
- #84
- #105

### Examples

User attempting to view an asset, but not signed in.
ScreenShot 2023-06-23 at 13 10 20@2x

User signed in, but doesn't have permissions to view an asset:
ScreenShot 2023-06-23 at 13 10 56@2x

User attempts to view an invalid asset. The ui route resolves, but the api request returns a 404:
ScreenShot 2023-06-23 at 13 13 21@2x

### Home page edge cases

When resolving errors, I found two cases where the home page seemed broken. 

1. The SignIn Notification would appear if a user didn't have search/browse permissions but were signed in. (e.g. a guest login).

I changed this to only show notification if a user wasn't signed in.

This led to:

2. The Home Page would be blank if a user was signed in, but they didn't have permissions to search/browse. This looked broken to me so I added a fallback notification, so the users knew that the page was intentionally left blank.

Though, I think I'm recalling some conversation about a few instances that want a blank home page for some reason. Let me know if I should pull out that bit, or hide it behind some other check.

ScreenShot 2023-06-23 at 13 00 11@2x


jxjj:
Also, I whole-heartedly welcome any copy-editing on the error message text.

cmcfadden:
I think just change "resource" to "asset" so that it's consistent? Maybe ditch the "Sorry." since we don't use that sort of humanized phrasing elsewhere?

jxjj:
We do have sorry's in other areas of the app, but I removed it here and elsewhere for consistency. (And I have no apologies about it.)

The error messages could show up on non-asset pages too: search results, drawers, static pages. Maybe we should just leave out the noun and just use a generic "this" as in "You do not have permission to access this." What do you think?

cmcfadden:
Sorry

I think a generic "this" is ok. 

Sorry

jxjj:
Done!

Issue #104, opened 2023-04-11
Multiselect widget is using `href` instead of `router-link`


jxjj:
It's doing a full refresh.

https://dev.elevator.umn.edu/dcl/asset/viewAsset/56c0272d2295a88565b640d3

jxjj:
This also doesn't properly set the advanced search filters after the click. Using the searchStore to perform the search and then doing a `router.push` to redirect should resolve both the full page load and the search filter issues.

Anything using `<ClickToSearchLink>` will probably have the same issues:

- [ ] MultiSelect Widget
- [ ] Date Widget
- [ ] Select Widget
- [ ] Text Widget

PR #180, opened 2023-06-23
Fix: multiselect widget does full reload on search instead of using vue-router

jxjj:
Part of:
- #104 

Previously, clicking on a multiselect widget within an asset would trigger a full page reload. Additionally, on the results load, the correct filters weren't set. This modifies links in the multiselect widget to do an advanced search when clicked, which resolves both issues.

The `<ClickToSearchLink>` component is left unmodified for now. As other widgets are using it (and probably have similar issues).

### Screenshots

Before: Advanced search value is for Greater Location field is blank
before

After: Multiselect for Greater Location is properly set
before

### DIFF
--- a/src/components/ClickToSearchLink/ClickToSearchLink.vue
+++ b/src/components/ClickToSearchLink/ClickToSearchLink.vue
@@ -1,11 +1,8 @@
 <template>
   <span v-if="props.widget.clickToSearch">
-    <button
-      class="text-blue-600 hover:text-blue-700 hover:underline"
-      @click="handleClick"
-    >
+    <a :href="linkUrl">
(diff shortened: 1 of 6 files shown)


cmcfadden (commented):
I think this breaks support for the clickToSearchType setting on the widget? Rather than using this approach, would it make more sense to refactor the querySearch and scopedQuerySearch endpoints to just return a JSON searchID? That way we're not supporting this extra logic in multiple places, and we don't need the custom multiSelectSearchLink component?

jxjj:
Sounds good.

jxjj:
Closing. Superseded by #181.

PR #181, opened 2023-06-27
Fix: Widgets using `<ClickToSearchLink>` component do a full page refresh

### LINKED ISSUE #104: Multiselect widget is using `href` instead of `router-link`

jxjj:
This resolves an issue where widgets using `<ClickToSearchLink>` would do a full page refresh on click. Now`<ClickToSearch>` will now grab the searchID when clicked, and redirect to the search. 

The Tag widget is changed to wrap `<Chip>` with `<ClickToSearch>`  rather than pass a plain href to the `<Chip>` component.

This also encodes search text in the URI, resolving an issue with some click to search text  (`Africa & Asia`).

- Complements UMN-LATIS/elevator#139
- Resolves #104
- Supersedes #180

### DIFF
--- a/src/api/fetchers.ts
+++ b/src/api/fetchers.ts
@@ -160,7 +160,8 @@ export async function fetchSearchIdForClickToSearch(
   widgetProps: WidgetProps
 ): Promise<string> {
   const url = toClickToSearchUrl(linkText, widgetProps);
-  const res = await axios.get<{ searchId: string }>(url);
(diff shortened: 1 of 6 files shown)


cmcfadden (approved):
looks good!

PR #184, opened 2023-06-29
Add asset to drawer from asset page

jxjj:
This adds the ability for a user to add an asset to a existing or new drawer from the asset page.

## Screenshots
ScreenShot 2023-06-29 at 14 40 39

ScreenShot 2023-06-29 at 15 05 20@2x

ScreenShot 2023-06-29 at 15 05 12@2x


cmcfadden:
Code looks good but I think we should think about placement. First, because I don't like stealing space from the asset, but also because it kind of implies that we're just adding the digital asset to the drawer and not the asset. Given that we know drawers are **very** low utilization, I think we can get away with making this less obvious?

jxjj:
Updates:
- Changed the placement of the add buttons so be with the bottom toolbar:
ScreenShot 2023-06-30 at 13 30 33@2x
- Having text for the Add to Canvas or Add to Wordpress plugin buttons looked weird, so changed to use similar iconography as Add to drawers:
ScreenShot 2023-06-30 at 13 31 18@2x
- changing the icon on success was too subtle, so now a toast will show up instead.

cmcfadden:
this is awesome, nice!

cmcfadden (approved):

PR #187, opened 2023-07-03
Add feature: Users can remove assets from drawers from the drawer results page

jxjj:
This adds a remove button to the assets listed on a drawer view page so that a user can remove the asset from the drawer:

ScreenShot 2023-07-03 at 17 40 22

Also fixes some cache-busting issues when mutating drawers.


cmcfadden (approved):
looks good

PR #188, opened 2023-07-06
Add to Drawer from Search Results Page

jxjj:
This allows a user to add an asset to a drawer from the search results page, the "More Like This" section, and the drawers page (in case a user wants to add an asset from one drawer to another).

## Screenshots

Add to Drawer button on search result card:

Workflow example:

Notes:
- There's some refactoring here since the AddToDrawerButton previously used the activeAssetId to determine which asset to add. On a search results page, there is no active asset id, so this must be passed as a prop.
- The Toast's confirmation message was too generic: "Added asset to MyDrawer drawer."  There are many assets on the search results page, so displaying a specific asset title is better. This adds some methods to the assetStore to get the asset title given an id. Also – I got tired of clicking back and forth between drawers while testing, so this adds a VIEW DRAWER button on the Toast if a user wants to quickly jump to the drawer after adding.
- The actions for adding an asset to a drawer were moved to the drawerStore so that the add to drawer button could be used on multiple pages.
- Both AddAssetToDrawerButton and the RemoveAssetFromDrawerButton.
- On the search results card, we want the `AddToDrawerButton` to not cause the whole care to highlight on hover and not cause navigation to the asset.  Instead of linking the whole card , this makes the image and the main card text area linked. The AddToDrawerButton is put in a card footer, which is not linked to the asset page.
- The RemoveFromDrawerButton is refactored to use the drawerStore actions, which has the nice side effect of the prop-drilling and event emitting.


cmcfadden:
Moving from a slack thread - let's not do the per-search-result add to drawer button. Old site lets you "Add to drawer" from a search results page which adds all of the current search results to a drawer. We can look at replicating that if it's not too much of a hassle?

jxjj:
Sounds good

jxjj:
Updates:
- Removed button on each card
- Added single button for adding all assets in search results.

ScreenShot 2023-07-05 at 23 30 40

cmcfadden (approved):
Looks good!

PR #191, opened 2023-07-07
add custom sort type

jxjj:
This adds sorting to drawers, including drag-and-drop custom sort. VueDraggable is added as a dependency for drag and drop magic.

Drawer browsing and sorting is very similar to normal search, but just different enough that I often just copied search components and then modified them slightly to work with drawers. Which is to say that there's a bit of duplicate code here that could probably use a refactor in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "vue": "^3.3.4",
     "vue-router": "4",
     "vue-toasted": "^1.1.28",
-    "vue3-popper": "^1.5.0"
+    "vue3-popper": "^1.5.0",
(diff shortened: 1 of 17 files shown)


cmcfadden (approved):
Code looks good - can we pin the height of drawer items like we do with search items, so that there isn't a shift when you drag between rows?

https://github.com/UMN-LATIS/elevator-ui/assets/211655/1840beb8-8676-4fb0-a47f-3f520201e520

jxjj:
Updated so that rows in drawer grid will be the same height:

ScreenShot 2023-07-07 at 09 12 15@2x

PR #193, opened 2023-07-10
add download drawer button and page

jxjj:
Adds a download button to the drawer page and a download page, which will kick off a download request –OR – if a download is already completed, it will show the download button.


src/pages/DrawerViewPage/DrawerViewPage.vue
```diff
@@ -238,11 +249,15 @@ const errorStore = useErrorStore();
...
+  const drawerRecord = await drawerStore.refreshDrawer(props.drawerId);
+
+  if (!drawerRecord) {
+    return errorStore.setError(new Error("Couldn't find drawer content."));
+  }
 
   // sortBy is persisted in the drawer record, so check its current value
-  const drawerRecord = drawerStore.drawerRecords[props.drawerId];
-  if (!drawerRecord.contents) {
+  // const drawerRecord = drawerStore.drawerRecords[props.drawerId];
```
cmcfadden, line 259: strip?

cmcfadden (approved):
looks good!

src/pages/DrawerViewPage/DrawerViewPage.vue
jxjj: ah, good catch. 

Adding this in while trying to debug a whiltescreen issue. Turned out my login had expired, and the drawer I was viewing wasn't accessible anymore, causing the page to break at `v-if="drawer.contents"`. Added a `v-if="drawer"` on the `<Tabs>`, which let the login prompt trigger.

All this is to say, keep your eye out for issues caused by login expirations.  If more bubble up, we should talk through some strategies for avoiding this.

PR #197, opened 2023-07-19
restore bottom border of app header

jxjj:
This reverts a change to the bottom border on the search result and drawer view pages based on feedback. Previously, the border was transparent to improve visability of the tabs by adding more contrasting whitespace around them. However, the appearing/disappearing app border when navigating could be distracting for some users.

Before: no bottom border on app header
ScreenShot 2023-07-19 at 12 50 29@2x 

After: bottom border on app header
ScreenShot 2023-07-19 at 12 50 10@2x


cmcfadden (approved):
looks good

PR #199, opened 2023-07-20
Add button for Editing Drawer Permissions, unify icon button styles

jxjj:
This adds an Edit Permissions button on the Drawer page and changes the download button to be a similar icon button.

There were a number of buttons that were roughly the same style as the ActiveFileViewButton, but not used on the FileViewPage. So, this PR renames the `<ActiveFileViewButton />` component as a more generic`<IconButton />` component.

The `<IconButton />` was modified to handle router links, hrefs, or plain old button clicks (similar to the `<Button>` component. Additionally, since it can be difficult to discern an icon's meaning, the IconButton now has a popover to show the button title on hover.

Then, icon buttons throughout the app were updated to use the <IconButton /> component.


cmcfadden (approved):
looks good

Issue #205, opened 2023-07-27
rename Create Drawer button ?


cmcfadden:
Should it be "Create Drawer and Add Item"? (In the old UI, it just created the drawer but left the modal up, new UI actually adds the item and dismisses the modal)

jxjj:
I thought about that and tried a few variations.

Here was the rationale:
- The longer button name made layout tricky and felt too long.
- I didn’t feel like more words on the button added clarity
- two clicks felt strange and confused me on the old site

In the end, context seemed enough to support a one click operation and shorter button names:

1. The modal is dismissed when clicked implying completion
2. The toast appearing confirms that it was added to the new drawer.

Whatever the case, the ui is changing to support excerpts. We can talk through adjustments in that PR.

Issue #133, opened 2023-05-15
Add Drawers to Elevator


jxjj:
A drawer is a user-made group of assets.  The most common use case is faculty creating a drawer for specific assets to use with a specific lecture, like slides.

Not all users can create drawers, only users with specific permissions.

### Functionality

- [x] #164
- [x] #165
- [x] #166
- [x] #167
- [x] #168
- [x] #169
- [x] #170
- [ ] #171
- [x] #172

### Out of scope

- Users cannot update a drawer name once created. This functionality doesn't currently exist.
- creating an excerpt page
- sharing, i.e. editing permissions of the drawer.
- previous/next functionality like search results
- recent drawers

cmcfadden:
Endpoints for drawers:

* List of all drawers user has: `/drawers/listDrawers/true`
* Contents of drawer: `/drawers/getDrawer/***$drawerId***`
* Remove item from drawer: `/drawers/removeFromDrawer/***$drawerId***/***$assetId***/true`
* Remove an excerpt from a drawer: `/drawers/removeExcerpt/***$drawerId***/***$excerptId***/true`
* Set sort order (sorting for drawer): `/drawers/setSortOrder/***$drawerId***/***$sortOrder***`
* Set custom sort order (manual dragging): `/drawers/setCustomOrder/***$drawerId***` (POST)

Custom sort ordering is probably a low priority, as that involves drag and drop and whatnot.

PR #206, opened 2023-07-29
Add excerpts to drawers

jxjj:
This allows users to add excerpts of video and audio files to drawers, and view the excerpts.

## Details
When a user clickes the "Add to Drawer" button, if the file is excerptable the modal will show a checkbox:
ScreenShot 2023-07-28 at 19 53 02@2x

When checked, the user can scrub to a location and click SET to set the start or end time, or they can enter the times in timestring format (seconds, MM:SS, HH:MM:SS, etc):
ScreenShot 2023-07-28 at 19 53 49@2x

The iframe within the modal is an`<ExcerptableIframe>` component, which handles sending/receiving messages parent window and `jwplayer` within iframe (e.g. updating the parent with the curentScrubberPosition)..

The Excerpt section does some basic input validation, if a user selects a drawer AND fills in text for creating a drawer:
ScreenShot 2023-07-28 at 20 30 05@2x

Or if they're missing some required data for the excerpt:

The "Add to Drawer" button is disabled until all fields are valid. When adding an excerpt, the excerpt must be at least 1 second long to be valid.

If, for some reason, the start time is after the end time, the submit function will swap the values so that they're in the correct order (I triggered a server error or playback issues if these values weren't valid).

When added to the drawer, an excerpt will show the correct excerpt label (not the asset title), and an "Excerpt" chip to differentiate it from other assets within the drawer.

The excerpt page layout is simple, showing the excerpt label, start/end times, a "View Asset" link, and buttons for downloading, sharing, and viewing details:
ScreenShot 2023-07-28 at 20 17 58@2x

Again, the excerpt page will load an `<ExcerptableIframe>`, which will send a message to the child iframe  to `setPlaybackBounds`.  I'm currently logging child/parent messages to the console. I'd like to leave this in for now to help with debugging.

The RemoveFromDrawer button was updated to handle removing excerpts, which is a slightly different api call. Confirm modals and toast were also updated to show the excerpt label and not the asset title:

ScreenShot 2023-07-28 at 20 24 02@2x

Finally, I added some subtle iconography to video and audio thumbnails, which made it a bit easier for me to tell which files should be excerptable:

Closes:
- #171 
- #133 
- and maybe #205 (there's no longer a Create Drawer button at all)


src/components/ThumbnailImage/ThumbnailImage.vue
```diff
@@ -15,27 +15,47 @@
...
+    fileType?: string | undefined;
   }>(),
   {
     href: undefined,
     isActive: false,
+    fileType: undefined,
   }
 );
+
+const isVideo = computed((): boolean => {
```
cmcfadden, line 50: If we’re going to use file extensions here, they should mirror the allow list types from audiohandler/moviehandler.

cmcfadden (commented):
One thing I found in testing - if you start the movie playing (in the main window) then open the drawer modal, you're getting events from both players so the times get confused. Maybe we can emit a pause when opening the popup

cmcfadden (approved):

jxjj:
Updates:
- file extensions for video and audio match those in elevator
- fixed iframe play/pause issues. I refactored a bit so that I could scope messages to a particular iframe. Now a pause will be sent to the main object viewer when the modal is opened. Similarly, if the modal is cancelled a pause is sent to the the excerpt iframe to prevent it from continuing to run.

On dev for review

cmcfadden (approved):

PR #211, opened 2023-08-08
fix invalid hash id error

### LINKED ISSUE #208: Clicking a related asset triggers "The string did not match the expected pattern."
This is caused by `router.ts` `scrollBehavior` trying to call ` document.querySelector(to.hash)`. Seems like it doesn't like the incoming hash in these cases ("#56a3bead7d58ae1f4b8b49d6").

Screenshot 2023-08-07 at 7 12 07 PM

jxjj:
`document.querySelector` will throw if `to.hash` isn't valid. e.g. if the id begins with a number instead of a letter. 

This switches to `getElementById` and only attempts the scrollTo behavior if the element is found.

Resolves #208

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -32,17 +32,23 @@ const router = createRouter({
       return savedPosition;
     }
 
-    // scroll to anchor if it exists on the page
-    if (to.hash && document.querySelector(to.hash)) {
+    const anchor = to.hash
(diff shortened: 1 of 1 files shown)


cmcfadden (approved):
... this feels like a better solution 😂

PR #227, opened 2023-08-28
Support pages with iframes

### LINKED ISSUE #225: Support pages with iframes
Example here:
https://dev.elevator.umn.edu/defaultinstance/page/view/21

Has an <iframe> embed, but it doesn't render

jxjj:
This permits embedded `<iframe>` tags within sanitized html content. Fixes #225.

See: https://dev.elevator.umn.edu/defaultinstance/page/view/21
ScreenShot 2023-08-28 at 14 40 42@2x


cmcfadden (approved):
Any idea what the set of allowed (or disallowed) tags is?

jxjj:
Allowed tags: https://github.com/cure53/DOMPurify/blob/main/src/tags.js
Allowed attrs: https://github.com/cure53/DOMPurify/blob/main/src/attrs.js

PR #228, opened 2023-08-28
navigate to next file or related asset with left/right keypresses

jxjj:
This adds keyboard shortcuts for navigating between related assets or uploads on an asset view page. When viewing a file or object that's active, pressing left or right will move to the next/prev item within the group.

ScreenShot 2023-08-28 at 16 25 33


cmcfadden (approved):
yayyy looks good

PR #235, opened 2023-10-02
fix: related asset titles not being used in some cases.

### LINKED ISSUE #234: Related Asset Cache Titles aren't being used in some cases
Sample asset:
https://dcl.elevator.umn.edu/asset/viewAsset/56e0267b7d58ae2e318babc5

New UI:
Screenshot 2023-09-29 at 10 24 28 PM

Old UI:
Screenshot 2023-09-29 at 10 24 35 PM

Values are being served in the relatedAssetCache (measurement for example):
```
        "56e0267b7d58ae2e318babc9": {
            "primaryHandler": null,
            "readyForDisplay": true,
            "relatedAssetTitle": [
                "59 m"
            ]
        },
```

I believe this is because RelatedAssetWidget.vue passes the assetCache in as `assetCacheItem` but AccordionRelatedAssetWidgetItem.vue is expecting it to be named `assetCache`? Didn't push a commit cause I might be misreading.

jxjj:
This fixes #234 : Related Asset Cache Titles aren't being used in some cases 

The suggestion issue was correct: The <RelatedAssetWidget> passes `assetCacheItem`, not `assetCache` to child components, but some of those children were expecting an `assetCache` prop. This updates to use `assetCacheItem`.

(Aside: it'd be nice if one could declare an interface for dynamic `<component>`s, so the linter would check if there's a prop mismatch like this... maybe there is?)

### DIFF
--- a/src/components/Widget/RelatedAssetWidget/AccordionRelatedAssetWidgetItem.vue
+++ b/src/components/Widget/RelatedAssetWidget/AccordionRelatedAssetWidgetItem.vue
@@ -1,14 +1,14 @@
 <template>
   <div class="accordion-related-asset-widget-item w-full max-w-lg">
     <Accordion :label="title">
-      <template v-if="assetCache?.primaryHandler" #label>
+      <template v-if="assetCacheItem?.primaryHandler" #label>
         <div
-          v-if="assetCache?.primaryHandler"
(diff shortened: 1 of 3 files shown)


jxjj:
Before:


After

PR #237, opened 2023-10-02
fix: lng lat label overflow

jxjj:
Using the `inline` variant of Tuple causes will crop off some of the lng data and the accordion chevron because "Longitude" and "Latitude" are a bit long.

This:
- abbreviates with "Lng" and "Lat"
- removes unnecessary right padding that's causing a flexbox overflow with the chevron.
- adds flex-wrap, just in cause some screensizes or fonts or whatever still cause an overflow.


cmcfadden (approved):
but what if I need to know which specific atom is being referred to?

PR #245, opened 2023-10-04
fix: map marker not rendering if no primaryHandlerId

### LINKED ISSUE #243: Not all results seem to be loading in collection maps
For example, on https://nibsda.elevator.umn.edu/search/s/ad41ae86-bf70-4453-b807-7a1a1278640c?resultsView=map, after loading all results in the new ui:

ScreenShot 2023-10-03 at 16 24 19@2x

But in the legacy ui:
ScreenShot 2023-10-03 at 16 26 44@2x

Note that there are a lot of results missing (and no results from alaska and hawaii)

jxjj:
This resolves #243, an issue where only a subset of results with locations where showing on the map search results page.

ScreenShot 2023-10-03 at 20 23 39@2x

The issue was the check for `primaryHandlerId`: when a search result had no `primaryHandlerId`, the loop skipped rendering the map marker.

A `primaryHandlerId` is needed for rendering a thumbURL, but not every match will have a thumbnail  (assets that are  "metadata only" won't have a `primaryHandlerId`, but they will still have a location). 

This PR allows `imgSrc` to be null, and instead skips any matches without locations.

An additional issue was uncovered once all map markers started loading: `Maximum recursive updates exceeded.`

This was caused by a conflict between the `watch`ers on `markers` (vue api) and the event listener for `styledata` changes (mapbox api).  I think the old `watch`ers on `markers` were before we started clustering map markers on zoom. Map clustering requires rending markers a different way. So, I've removed the old `watch` on the markers. (I didn't see a noticeable change in performance, but it stopped the vue warnings.)

### DIFF
--- a/src/components/Map/Map.vue
+++ b/src/components/Map/Map.vue
@@ -65,6 +65,7 @@ import { LngLat, BoundingBox, MapContext, AddMarkerArgs } from "@/types";
 import { MapInjectionKey } from "@/constants/mapConstants";
 import Skeleton from "../Skeleton/Skeleton.vue";
 import { Point } from "geojson";
+import { useThrottleFn } from "@vueuse/core";
 
 const props = withDefaults(
   defineProps<{
@@ -176,7 +177,7 @@ function getOtherMarkersWithSameCoords({
   });
 }
 
-function createOrUpdateMarker({ id, lng, lat, ...properties }: AddMarkerArgs) {
+function addMarker({ id, lng, lat, ...properties }: AddMarkerArgs) {
   const otherMarkersWithSameCoords = getOtherMarkersWithSameCoords({
     id,
     lng,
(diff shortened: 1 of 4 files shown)


src/components/Map/Map.vue
```diff
@@ -176,7 +176,7 @@ function getOtherMarkersWithSameCoords({
   });
 }
 
-function createOrUpdateMarker({ id, lng, lat, ...properties }: AddMarkerArgs) {
+function addMarker({ id, lng, lat, ...properties }: AddMarkerArgs) {
```
cmcfadden, line 179: What's the thinking on changing this name?

src/components/Map/Map.vue
```diff
@@ -380,7 +377,7 @@ onMounted(() => {
     .on("mouseleave", UNCLUSTERED_LAYER_ID, function () {
       map.getCanvas().style.cursor = "";
     })
-    .on("styledata", () => {
+    .on("styledata", (event) => {
```
cmcfadden, line 380: Are you consuming this somewhere or is adding the param just for future clarity?

cmcfadden (approved):

src/components/Map/Map.vue
jxjj: I stopped `watch`ing the marker lng/lat, and switched to `onMounted`, so marker updates won't percolate through.

src/components/Map/Map.vue
jxjj: Good spot. I was logging it out when troubleshooting to see what was triggering the styledata events. I'll remove.

PR #253, opened 2023-10-20
Add some helpers for editing static content pages

jxjj:
- Adds layout classes that content editors can use on static pages if they want to add some layout. These are namespaced under `el-` to keep these abstracted away from a specific framework (tailwind, bootstrap). I chose a BEM approach where multiple classes can be applied like `el-grid el-grid--2/3-1/3` for a 2 column layout with first col 2/3 width and second 1/3 width.
- NIBSDA static page content has been prepped for migration to new ui. They're available for review on https://dev.elevator.umn.edu/defaultinstance/. Content is also saved in [its own private repo](https://github.com/UMN-LATIS/nibsda-static-page-content) with a few notes. Principally, the content just needs to use semantic markup (`p` instead of `h5`), remove inline styles, and change to `el-` classes where advanced layout is needed.
- Add an "Edit" button to static content page for Admins:

Edit button added to page

- fixed some alignment on the static pages. Content container and page title now centered.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,1 +1,4 @@
-{}
+{
+  "htmlWhitespaceSensitivity": "ignore",
+  "bracketSameLine": true
+}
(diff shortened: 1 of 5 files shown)


cmcfadden (approved):
looks great

PR #254, opened 2023-11-13
Make search results map, gallery, and timeline view embeddable

### LINKED ISSUE #240: Add map, timeline and gallery embedding
(mirror existing site)

Embeds should track UI setting for site.

jxjj:
ScreenShot 2023-11-13 at 16 14 46@2x

- adds an share button to map, gallery, and timeline tabs of search results page
- adds route for embed views

Closes #240 

Question: How do we handle the number of results loaded? Should there be a load more button on the embed views? Or some way to ensure the same results are loaded for embed?

### DIFF
--- a/src/api/fetchers.ts
+++ b/src/api/fetchers.ts
@@ -246,11 +246,23 @@ export async function fetchSearchId(
 
 export async function fetchSearchResultsById(
   searchId: string,
-  page: number,
-  loadAll: boolean
+  opts: {
+    page?: number;
+    loadAll?: boolean;
+  } = {}
 ) {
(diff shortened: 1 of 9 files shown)


cmcfadden (approved):

cmcfadden:
the current version just calls the loadall every time, which is probably our best bet?

jxjj:
Updates:
-  loadAll on map and timeline embeds. Gallery still lazy loads.
- fixed console warning about props passed to share button. the button had two direct descendants, since it had a modal – which meant vue didn't know which child to attach the `class` attribute to.

PR #261, opened 2024-02-01
Feature/lti13support

cmcfadden:
Adds handling for LTI13. Basically this just involves posting to a different endpoint and including some additional info - the methodology is the same as lti1.1.

### DIFF
--- a/src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
+++ b/src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
@@ -53,6 +53,7 @@ const {
   returnUrl,
   elevatorLTIVersion,
   elevatorLaunchId,
+  userId,
   clear: clearElevatorSessionStorage,
 } = useElevatorSessionStorage();
 
@@ -121,6 +122,7 @@ async function onConfirmedToAdd() {
         excerptId: "",
         returnUrl: returnUrl.value ?? "",
         launchId: elevatorLaunchId.value ?? "",
+        userId: userId.value ?? "",
       });
 
       document.body.innerHTML += data;
--- a/src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
+++ b/src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
@@ -23,7 +23,7 @@
     <p v-else>Are you sure you want to add this to {{ elevatorPlugin }}?</p>
   </ConfirmModal>
   <form
-    v-if="elevatorCallbackType == 'lti'"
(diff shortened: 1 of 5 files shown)


src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
```diff
@@ -111,24 +114,39 @@ async function onConfirmedToAdd() {
...
+      const data = await api.postLtiPayload13({
+        fileObjectId: props.fileHandlerId,
+        excerptId: "",
+        returnUrl: returnUrl.value ?? "",
+        launchId: elevatorLaunchId.value ?? "",
+        userId: userId.value ?? "",
+      });
+
+      document.body.innerHTML += data;
+      (document.getElementById('auto_submit') as HTMLFormElement)?.submit();
```
jxjj, line 129: Sorry – I'm not following this part. What's happening here? Is this something specific to how LTI1.3 works?

src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
```diff
@@ -111,24 +114,39 @@ async function onConfirmedToAdd() {
...
+      nextTick(() => {
+        if (!returnForm.value) {
+          addingToPluginStatus.value = "error";
+          throw new Error("Return form not set");
+        }
+
+        returnForm.value.submit();
+        addingToPluginStatus.value = "success";
+        toastStore.addToast({ message: "Added to Canvas" });
+      });
```
jxjj, line 148: Should we show a success message to the user in LTI1.3 as well? Or maybe this isn't required because they'll see something from the `document.body.innerHTML += data` line above?

jxjj: nevermind!

jxjj:
lgtm! flow works great. 

Just my one question about the form autosubmit thing – you could maybe add a comment about the intent of those lines?

jxjj (approved):

src/components/AddToEmbeddedPluginButton/AddToEmbeddedPluginButton.vue
cmcfadden: yeah, more or less - because DeepLinking happens inside an iframe, we need to get the client itself to make the deep linking submission to the LMS (so the LMS can then close the iframe/do whatever). So it can't be server<->server like grade passback is with LTI1.3. 

The specific auto_submit ID name comes from the LTI library we use.

jxjj: ah, that makes sense. Also why I couldn't find the `auto_submit` id anywhere in the elevator or ui code. Thanks!

PR #278, opened 2024-11-06
fix: add asset sometimes errors

### LINKED ISSUE #277: Add Asset sometimes generates on error
there’s an issue i see sometimes on elevator dev with the new ui

1. login with a local user
2. bring up menu to add an asset
3. select a template, click submit
3. land at an error page about unknown template

Maybe what happens is that when you hit submit, you’re falling into the legacy interface, and then it redirects you through shibboleth to see if it can bring up a passive elevator session so if you try again, it works, because now you have that cookie set

jxjj:
On dev. LMK if there's a better approach or if it interferes with some other functionality.

resolves #277

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "dompurify": "^3.0.3",
     "focus-trap": "^7.4.3",
     "geojson": "^0.5.0",
+    "js-cookie": "^3.0.5",
     "mapbox-gl-esri-sources": "^0.0.5",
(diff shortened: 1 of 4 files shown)


cmcfadden:
ah sorry, didn't know you were working on this. I fixed this on the server side a few months ago, in 54999af0e415b1aaa97570d80def92cc85b151c7. Just didn't remember this issue was open.

jxjj:
good enough!

PR #281, opened 2024-12-12
fix: wordpress plugin not initializing when elevatorPlugin window o…

### LINKED ISSUE #256: Wordpress plugin not working
I can't recall if this worked at one point or not?

The Wordpress plugin works by spawning a window with the name 'elevatorPlugin'. When Elevator loads, it posts a `parentLoaded` message back to the opener, which responds with the rest of the Elevator plugin payload.

jxjj:
Fixes #256 

- adds wordpress plugin initialiation logic if `window.name` is `elevatorPlugin`.
- adds some fixes for missing types.

https://github.com/user-attachments/assets/a06eb62e-8134-46ec-bb70-487dab8903df

## Details

I question whether the wordpress plugin was actually working correctly in the vue template before. I couldn't find ANY place in the git history where we were triggering the `parentLoaded` message or setting the client-side session storage vars. I wonder if I was triggering setup logic when toggling back/forth between classic and vue mode?

Now:
1. when the app is mounted, we're checking for`window.name === "elevatorPlugin`". If so, we:
2. setup the event handler to listen for a message from the parent window with initialization info.
3. send the `parentLoaded` message to the parent.

I don't _think_ any of these changes impact LTI 1.3 stuff, but would really appreciate a second set of eyes. The fact that I'm even questioning makes me wonder if I should split the WP stuff and the LTI stuff into independent components?

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,5 +91,6 @@
     "typescript": "^5.0.4",
     "vite": "^4.3.9",
     "vue-tsc": "^1.7.8"
-  }
+  },
+  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
 }
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,10 +8,11 @@
   </div>
 </template>
 <script setup lang="ts">
-import { onMounted } from "vue";
+import { onMounted, onUnmounted } from "vue";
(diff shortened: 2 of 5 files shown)


cmcfadden (approved):
I think this seems right. I'll do a separate pass of the LTI stuff once it's merged - I have a few other changes in mind for that over the break.

PR #286, opened 2025-02-14
better handling of long app names

jxjj:
Handles long app names better by adding a min-width to the advanced search and a max-width to app-name.

https://github.com/user-attachments/assets/1bd2ffd3-19b9-4684-b48e-1dd8f1c3912c

not yet on dev


cmcfadden (approved):
nice

PR #287, opened 2025-02-14
add a boolean toggle to expose some ES functionality

cmcfadden:
(edit: image didn't stick)

Screenshot 2025-02-14 at 6 34 48 PM

### DIFF
--- a/src/__mocks__/mockSearchResults.ts
+++ b/src/__mocks__/mockSearchResults.ts
@@ -806,6 +806,7 @@ const mockSearch: SearchResultsResponse = {
   success: true,
   searchEntry: {
     showHidden: "0",
+    useBoolean: "0",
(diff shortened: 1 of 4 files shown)


jxjj (approved):
Yes AND yes!

Seems super useful. Are there cases when this shouldn't be used? Could it just be on all the time (or whenever an "AND" or "OR" appears in a query), or would that make it harder to find some things?

Nit: The alignment seems a little off to my eye. I wonder if it would look more in place if aligned with the other inputs.

ScreenShot 2025-02-18 at 09 39 03@2x

cmcfadden:
I'll tweak the spacing.

The search tool that elastic uses which has boolean capability isn't as good for a normal user - it doesn't have much flexibility around natural language, field boosting, etc. It's good for library nerds.

PR #289, opened 2025-02-25
test tracking asset view with gtag

jxjj:
- Adds a composable `useAnalytics` which can be used to send `view_asset` and `download` custom events to analytics. 
- dispatches custom events when an asset is viewed or a download is clicked.

The `view_asset` payload will include: 
- `asset_id`, 
- `asset_name`, 
- `collection_id`
- `collection_name`

`download` includes:
- `collection_id`
- `collection_name`
- `asset_id`
- `asset_name`
- `file_object_id`
- `file_type`

View asset event in tag assistant

download event in tag assistant

On dev for testing.

Other notes: 
- for reporting to work, payload properties will need to be set up as "custom dimensions" in the Analytics interface (this is done on dev). 
- No need to set up custom events in the analytics interface.
- There's a lag before the data is available for reporting, so we may need to tweak some things. But the events can be viewed using `https://tagassistant.google.com/`

### DIFF
--- a/src/helpers/useAnalytics.ts
+++ b/src/helpers/useAnalytics.ts
@@ -7,9 +7,12 @@ const VIEW_ASSET_EVENT = "view_asset";
 const DOWNLOAD_EVENT = "download";
 
 async function getAssetDetails(assetId: string): Promise<{
-  assetName: string;
-  collectionId: AssetCollection["id"];
-  collectionName: string;
+  asset_id: string;
+  asset_name: string;
+  collection_id: AssetCollection["id"];
+  collection_name: string;
+  instance_id: number;
+  instance_name: string;
 }> {
   const assetStore = useAssetStore();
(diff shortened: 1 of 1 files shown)


cmcfadden:
Is any of this showing in GA for you yet, or is that what you meant by the delay? It looks good in tag manager - would be nice to get instance name in, but I'm assuming that's on your roadmap.

jxjj:
heh – "roadmap." Good catch with instance name. I was thinking GA property = instance, but that's probably not true. Any other data that'd be helpful to grab?

> Is any of this showing in GA for you yet

It's showing up in the Realtime Overview events panel (below), but not the reports yet. Yes, that's what I meant by the delay... so it's hard to know yet if things are working fully or not.

ScreenShot 2025-02-25 at 18 26 27@2x

PR #293, opened 2025-03-04
Add a Download All button for downloading all asset files

### LINKED ISSUE #282: Explore multi-file downloads
Could we allow a user to download all the digital assets attached to a record? I've seen websites do that but I don't know if there's actually an API for that or if it's just javascript window.locations with a bunch of timers?

jxjj:
Closes #282 

- Adds a download all button the Uploads Widget.
- when clicked, we'll get download information for every file and then start a download stream for each. The streams are piped to `downloadZip()`.
- adds `client-zip`, a very nice library for combining multiple streams into a single zip file.

Tried a few other approaches, but ran into issues with each. (e.g. max number of concurrent downloads, popup blockers, trying do download 200 files and getting 200 prompts for where to save and what to name files. I had to quit my browser with that one...)

Let me know if you want me to move the button. I can avoid a component refactor if I put it where it is... but it probably looks better moved up on the same line.

On dev for testing.

ScreenShot 2025-03-03 at 20 57 13@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "react-dom": "^18.2.0",
     "shave": "^5.0.2",
     "storybook-vue3-router": "^2.3.1",
+    "streamsaver": "^2.0.6",
     "striptags": "^4.0.0-alpha.4",
     "swiper": "^9.3.2",
     "tiny-invariant": "^1.3.3",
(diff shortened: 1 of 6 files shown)


cmcfadden:
let's punt on this until we can review it synchronously or in person.

jxjj:
Sounds good.

One change I made: I realized that I wasn't streaming the zip stream to disk, and instead I was loading it into memory with `.blob()`. That should be resolved now.

jxjj:
Closing. Superseded by #296 after discussion.

PR #295, opened 2025-03-10
fix: white screen when analytics key undefined (early return)

jxjj:
An alternative approach if `noop` is too "clever" (in a bad way).


cmcfadden (approved):
I think I prefer this for the clarity

PR #296, opened 2025-03-11
Add dropdown menu options for downloading all assets in upload widget

jxjj:
Closes https://github.com/UMN-LATIS/elevator-ui/issues/282

This supersedes #293 

- Adds a "more actions" menu to the Upload Widget component to allow users to download all derivatives (or originals, if permitted).
- Some browsers (safari) it seems will abort any in-flight download requests if a response hasn't been returned from the server, so an artificial timeout is added ~2s to wait for the server to respond. (The download doesn't need to finish within 2s, it just needs to start within that time or an error will be thrown).
- To determine whether or not a user can download originals, we need to get file info for the first download. Since using the download all feature is rare, we avoid doing with every upload widget rendering. Instead, the permission check will happen once the user attempts to open the "More actions" menu. It's a little hacky.

On dev for testing.

https://github.com/user-attachments/assets/91eb2510-a6d1-4246-b86e-7c220ded20c6


cmcfadden (approved):
looks great

PR #300, opened 2025-03-17
fix: search sort sometimes listed incorrectly

### LINKED ISSUE #298: Search sort is funny in new ui
- Sometimes "Best Match" isn't properly showing as selected in search results.

jxjj:
Resolves #298 

Sometimes, the `sort` field in UI wouldn't match the search results. For example, if you typed "test" in the search input. The results returned would have `searchEntry.sort: 0` (best match), but the search selector would show `Default Title` 

The cause:
- On initial page load, the query is empty as the page waits for the search results to load.
- Previously, we had logic to set an initial sort state. It would use a query param (if one existed), or "Best Match", or – and this is where the problem is – it would use "Default Title" if the search was blank (i.e. we were asking for all results).
- Because the query was empty on page load (since data wasn't loaded), the initial sort was "Default Title"
- THEN – and this is error #2 – we were not properly updating the sort  in `searchStore` with the response from the server.

This:
- fixes searchStore to properly update sort once results are received.
- removes the logic in SearchResultsPage to set an initial sort while waiting for results
- removes logic to change "Best Match" to "Default Title" on empty search. The server response seems correct, so we shouldn't need this if we don't bother with setting initial sort before the results arrive.

### DIFF
--- a/src/pages/SearchResultsPage/SearchResultsPage.vue
+++ b/src/pages/SearchResultsPage/SearchResultsPage.vue
@@ -223,32 +223,9 @@ async function handleSortOptionChange(sortOption: keyof SearchSortOptions) {
   });
 }
 
-function getInitialSortOption(searchQuery: string): keyof SearchSortOptions {
-  const sortOption = route.query.sort as keyof SearchSortOptions;
-
-  // if the query is blank and the sort option is best match
-  // then we should default to sorting by title, since best match
-  // doesn't make sense for a blank query
-  if (
-    searchQuery === "" &&
-    (!sortOption || sortOption === SORT_KEYS.BEST_MATCH)
-  ) {
-    return SORT_KEYS.TITLE;
-  }
-
-  // otherwise return the sort option from the query param
-  // or default to best match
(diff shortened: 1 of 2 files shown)


cmcfadden:
I can still reproduce an issue with these steps:

- from main page, hit an empty search
- from search results page, hit the "dev default" in the upper left to go back to the main page
- do a search ("all files")
-> the resulting search is sorted by "default title"

I'd also like to explore what would be involved in some logic to replicate the current behavior from the old UI, which would be a special case for a blank search where we don't persist the sorting. A real world example of why this matters:

- browse to a collection (which effectively does an empty search filtered to that collection)
- go "oh gosh this is a lot of stuff!" and enter a specific search term
- results are sorted by title, instead of best match - that seems like a bad state to be in

In the old UI, that process would end up sorted by best match.

jxjj:
The reproduction is likely because sort is "sticky". The first empty search is Default Title, and a subsequent search will be too unless changed. 

Changing the empty search behavior as suggested should resolve this. Also prob less surprising behavior.

Don't think it should be too hard

cmcfadden:
this looks good to me

PR #316, opened 2025-07-29
fix: transitions from create asset to edit asset

jxjj:
- fixes page would flash for reals 🤞
- fixes "Create Asset" appearing on "Edit Asset" page.
- Add second "All Fields" template to mock server for testing.

On dev for testing.


cmcfadden (approved):
nice!

PR #326, opened 2025-08-04
Debounce asset editor saves

jxjj:
This supersedes #321 

- saves uploads when the begin, not just when they complete to avoid orphaned uploads.
- debounce save emits to cut down on requests (~2s, max: 5s)
- adds test for uploading multiple assets
- fixes a warning when submitting form (handler was attached to button, not the form's submit event)
- block arcgis requests when testing
- refactor some useAssetEditor methods to use promises instead of callbacks
- renames some functions so they're more distinct from in-component handler names


cmcfadden (approved):
cool cool

PR #330, opened 2025-08-12
inline related asset widget editing

jxjj:
This implements inline related asset editing, allowing users to create/edit related assets directly within parent assets.
 
- moves related asset widget stuff to its own `EditRelatedAssetWidget` folder.
- adds `EditRelatedAssetInlineContentItem` component for inline asset editing within widgets
- adds `InlineCreateOrEditAssetPage` view for editing assets inline assets.
- Inline assets register an `onBeforeSave` hook with the parent asset editor to trigger the inline asset saves before parent asset saves
- inline assets begin with any non-required empty fields collapsed to reduce visual complexity
- added unique editor instance IDs to prevent form collisions. With inline assets there could be multiple widgets on the page with the same widget. This would cause vue to update multiple inline fields with the same data since the key was not unique.
- toast notifications for save success/error states
- testing stuff

on dev for testing


cmcfadden (commented):
can look through this more during our meeting today, but a couple things:
1) not sure if it was broken by this PR, but the defaultTemplate for non-inline related assets seems to be broken now?
2) the toast covers up the "view" button so after clicking save you have to wait for it to clear before you can view your asset

PR #335, opened 2025-08-14
fix: update primary file handler id after save

jxjj:
After saving, the primary file handler id may need to be updated. This fetches the asset after saving, and then updates primary file hander id (and a few other fields).


cmcfadden (approved):
lg

PR #368, opened 2025-09-10
Collection list for add/edit should be "collections I can edit" not "collections I can view"

cmcfadden:
Just opening this so I don't forget about it. We can review when we're both back


cmcfadden:
We could, but we'd also need another key for visible, because there might be editable collections that aren't visible. That seemed like it introduced some extra complexity, because we don't want to risk leaking collection visibility to users who can't edit? But I'm fine with that change if it seems safe

jxjj:
Superseded by #370

PR #370, opened 2025-09-15
use collection `canView` and `canEdit` privileges for more fine-grained control of collection visibility

jxjj:
This complements  https://github.com/UMN-LATIS/elevator/pull/180 and builds on #368 . (Separate branch/PR, because I could be missing edge cases, which might make the other PR better).

If the user does not have edit permissions for ANY collection, there should be no change in what's rendered.

If they `canEdit` some collection, then the current business logic is:
- `AllCollectionsPage` - only show viewable collections
- `FilterByCollectionSection` - in advanced search, show but disable collections that can't be viewed (matches legacy behavior)
- `AssetEditor` - in collection section, show but disable collections that can't be edited (matches legacy behavior)

on dev.

TODO: fix tests

### DIFF
--- a/mock-server/db/collections.ts
+++ b/mock-server/db/collections.ts
@@ -8,6 +8,9 @@ const collectionSeeds: AssetCollection[] = [
     previewImageId: "",
     children: [],
     parentId: null,
+    showInBrowse: true,
+    canView: true,
+    canEdit: true,
   },
   {
     id: 2,
@@ -15,6 +18,9 @@ const collectionSeeds: AssetCollection[] = [
     previewImageId: "",
     children: [],
     parentId: null,
+    showInBrowse: true,
+    canView: true,
+    canEdit: false,
   },
 ];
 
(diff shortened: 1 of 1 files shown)


src/components/AdvancedSearchForm/AdvSearchDropDownItem.vue
```diff
@@ -1,9 +1,16 @@
 <template>
-  <DropDownItem class="!whitespace-nowrap overflow-ellipsis overflow-x-hidden">
+  <DropDownItem
```
cmcfadden, line 2: There's probably a couple things we want to do here. I think there's a bug in the old UI where collections not marked as browsable still show in the advanced search list. 

We also never want to reveal the existence of a collection that a user doesn't have permission to view. So we don't want to disable those, they just shouldn't exist as far as that user is concerned.

So this list should basically be a mirror the "list collections" page - only collections which are "browsable" and which the user has view permissions to, I think. So that'd basically mirror what we show on the list collections page?

PR #375, opened 2025-09-18
fix: some text area widget styling doesn't render

### LINKED ISSUE #369: Some text area styling doesn't carry through
Using things like the ordered list in Quill doesn't actually result in an ordered list on the display side.

Image

Image

jxjj:
This resolve several text area widget editor and display issues:

- Changed TextAreaWidget use CSS-based truncation (`line-clamp` with `max-height` fallback) instead of the `shave` library to resolve the HTML stripping that happened when the text area was collapsed.
- Added `prose` class on TextAreaWidget for better html formatting.
- The editor will clean-up Quill's HTML on save: quill-flavored HTML is converted to semantic html, empty paragraphs and prolific `&nbsp;` will be removed.
- Added CSS for handle Quill's HTML bulleted list output, just in case some non-semantic quill html exists in the wild.

Closes #369


cmcfadden:
Does this break the "Text Area Widget Collapsed Height (px):" configuration option in instance settings?

jxjj:
fixed

cmcfadden (approved):
lgtm

PR #396, opened 2025-12-09
emit events on static page content load and images load for custom scripts

jxjj:
This adds functionality for Elevator admins to add custom scripts which  can respond to page content loads or image loads.

A large chunk of the PR is adding tests to convince myself that we're correctly handling edge cases.

## Problem: Legacy App vs SPA Script Execution

As an example, suppose an admin wishes to add a custom script which turns  images on a particular static page into a carousel.

In the legacy UI, admins could add the script directly in the custom app header which adds an event  listener on load. Each new page navigation would trigger a full page  reload, rerunning all scripts and attaching event handlers as needed. This simple model worked because:
- Every navigation was a fresh start with a clean DOM
- Scripts could safely assume they ran once per page view
- Standard browser events (DOMContentLoaded, load) fired reliably

But the new UI is a single page app (SPA), which breaks this approach:

1. **Standard browser events don't fire on navigation.** Events like  `DOMContentLoaded` and `window.load` only fire once when the SPA initially loads, so scripts listening for these events will never respond to route changes.

2. **We cannot naively rerun custom scripts on each route change.** Each page nav isn't a fresh start with a clean dom in SPA navigation,  so re-execution of scripts may attach new event listeners or initialize components,  creating multiple handlers for the same elements. This leads to:
   - Memory leaks from accumulated listeners
   - Multiple responses to a single event (e.g., carousel initializing 3x)
   - Unpredictable behavior as state accumulates across navigations

3. **Custom scripts lack visibility into SPA routing.** Admin-provided scripts have no easy way to detect when the SPA has navigated to a new page. Monitoring route changes is not enough since the SPA will need to fetch page content from the api before it can update the DOM.

## Solution: Parse-Once Execution + Custom Lifecycle Events

This PR updates script handling to address these problems:

### 1. Parse and run custom scripts at app initialization (not when CustomAppHeader renders)

Previously, scripts were parsed in the `<SaniitizeHTML>` component as part of `<CustomAppHeader>`. This had two issues:
   - not every page has the custom app header, so navigation may cause a rerender and thus re-execution of these scripts
   - the default behavior of the sanitize component was to execute any embedded scripts, which isn't great

Instead, 
- we moved script parsing and execution to the `instanceStore` which is responsible for fetching and parsing the customAppHeader and footer already. 
- `instanceStore` now has a new property,`customScripts`, to hold any script elements from the `customHeader` and `customFooter`
- `instanceStore.customHeader`, `instanceStore.customFooter` properties will only hold scriptless HTML
- `<SanitizeHTML>` updated to remove scripts from html
- `instanceStore` will execute any `customScripts` once at app initialization

### 2. Emit custom lifecycle events on navigation:
   - `elevator:static-page:content-loaded` - Fired when new page content has been inserted into the DOM
   - `elevator:static-page:images-loaded` - Fired when all images on the new page have finished loading

Scripts can listen for these events to respond to navigation:
   
   ```javascript
   // Example: Initialize carousel on every page navigation
   document.addEventListener('elevator:static-page:images-loaded', () => {
     const carousel = document.querySelector('.carousel');
     if (carousel) {
       initCarousel(carousel);
     }
   });
```

### 3. Homepage feature assets

If the homepage has a feature asset, we wait for BOTH the page content and featured asset info to load before emitting events.

## Other Details
- we emit `images-loaded` if an image loads OR errors (e.g. 404)
- if a page has no images, we still emit `images-loaded`. 
- we set a timeout of 10s for image loads. If some image hasn't loaded within 10s, we consider it an error and fire the `images-loaded` event. We may want to change this behavior.
- for `content-loaded', the`event.detail` payload includes `{ pageId: 2 }` in case there's page specific JS.
- similarly, for `images-loaded`, the `event.detail` payload includes `{ pageId: 3, images: [...image elements] }`

## Testing

You can test the behavior on dev, by adding a custom script to the Custom Header admin setting:

```html
<script>
window.addEventListener('elevator:static-content-page:content-loaded', (event) => {
  console.log('Static content page loaded:', {...event.detail});
});

window.addEventListener('elevator:static-content-page:images-loaded', (event) => {
  console.log('Images loaded', { ...event.detail })
});
</script>
```

It should behave something like this:

https://github.com/user-attachments/assets/29be2ca5-e18c-4e5c-a1c8-6c40ead2ec49

### DIFF
--- a/mock-server/db/index.ts
+++ b/mock-server/db/index.ts
@@ -13,16 +13,17 @@ import { createUploadsTable } from "./uploads";
 const makeDb = () => {
   const collections = createCollectionsTable();
   const templates = createTemplatesTable();
+  const pages = createPagesTable();
   const assets = createAssetsTable({ collections, templates });
   const tables = {
     collections,
     templates,
     assets,
+    pages,
     searches: createSearchesTable({ assets, collections, templates }),
     users: createUsersTable(),
     sessions: createSessionsTable(),
-    instances: createInstancesTable(),
-    pages: createPagesTable(),
+    instances: createInstancesTable({ pages }),
     files: createFilesTable(),
     drawers: createDrawersTable(),
     uploads: createUploadsTable(),
--- a/mock-server/db/instances.ts
+++ b/mock-server/db/instances.ts
@@ -41,7 +41,13 @@ export function createInstancesTable({ pages }: { pages: PagesTable }) {
       if (baseTable.size() === 0) {
         throw new Error("No instances available");
       }
-      return instanceSeeds[0]; // Return the first/default instance
+      // Get the instance from the store (which may have been updated)
+      // rather than returning the seed data directly
+      const instance = baseTable.get(instanceSeeds[0].id);
+      if (!instance) {
+        throw new Error("Default instance not found");
+      }
+      return instance;
     },
   };
 }
--- a/mock-server/db/instances.ts
+++ b/mock-server/db/instances.ts
@@ -1,6 +1,7 @@
 import { ShowCustomHeaderMode } from "../../src/types";
-import { MockInstance } from "../types";
+import { MockInstance, MockPage } from "../types";
 import { createBaseTable } from "./baseTable";
+import { PagesTable } from "./pages";
 
 const instanceSeeds: MockInstance[] = [
   {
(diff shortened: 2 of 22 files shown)


cmcfadden (approved):
looks good

PR #406, opened 2025-12-19
Add spidering when points overlap

### LINKED ISSUE #400: Many items at the same lat/lon don't render correctly in the search results map
See this example, new UI / old UI. This collection has six items with the same lat/lon, but they don't seem to render quite right - only two pins show up no matter how far you zoom

Image

Image

jxjj:
- When points overlap, they spider instead of cluster
- Spidering will happen automatically on zoom
- Adds a test page for testing  http://localhost:5173/tests/map

Resolves #400

On dev for testing.

https://github.com/user-attachments/assets/dded2a44-1187-4b5b-9494-8a94b0ea9054

### DIFF
--- a/src/pages/TestPages/MapClusterTest.vue
+++ b/src/pages/TestPages/MapClusterTest.vue
@@ -1,5 +1,21 @@
 <template>
   <div class="map-cluster-test p-8">
+    <div class="mb-6 p-4 bg-gray-100 rounded">
+      <p class="text-sm font-semibold mb-2">Test Pages:</p>
+      <div class="flex gap-4 text-sm">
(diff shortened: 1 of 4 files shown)


cmcfadden:
It seems like there's a race in here - sometimes loading this page directly I get just gray, but I can zoom out and see the (5) point, then zoom in to the spider. No errors in console. Happens in both safari and chrome 
https://dev.elevator.umn.edu/defaultinstance/search/s/517b243c-129d-4575-aa08-a75bcc0b6167?resultsView=map

https://github.com/user-attachments/assets/80643ab8-b179-46ae-b30e-39a48f1e3e5c

PR #408, opened 2026-01-20
reset min-height to initial value on md+ screen size

### LINKED ISSUE #407: Rendering glitch when disclosing more info
See screenshot

Image

jxjj:
This resolves an issue where AssetDetailsPanel layout causes the ActiveFileViewToolbar to overlapping the ObjectViewer on `md+` screensizes. This explicitly unsets `min-height` (was `50dvh` previous), so that absolutely positioned items get the proper dimensions.

On dev.

Resolves #407


cmcfadden (approved):
good with me assuming copilots feedback isn't valid

PR #459, opened 2026-03-09
Add CSS classes so they can be used for hiding remote or local login options

jxjj:
This should do the trick for hiding local login.

```css
.sign-in-required__local-login,
.auth-drop-down__local-login { 
  display: none;
}
```
ScreenShot 2026-03-09 at 15 25 18@2x


cmcfadden (approved):
thanks!

PR #492, opened 2026-03-21
add Trash tab with delete and restore to All My Assets page

### LINKED ISSUE #468: Show deleted status
I don't think right now the editor shows assets that are in deleted status?

jxjj:
https://github.com/user-attachments/assets/872eaf54-3dc1-45dd-81e9-fceb1374816d

Resolves #468 

Complements endpoint changes in https://github.com/UMN-LATIS/elevator/pull/220

- Adds "My Assets" and "Trash (n)" tabs to the All My Assets page
- Trash tab lists the current user's soft-deleted assets with title, ID, and deleted date
- One-click Restore undeletes assets via `POST /assetManager/undeleteAsset/:assetId`
- Delete and restore both show toasts with asset title and Undo action
- Active tab persists in URL query param (`?tab=trash`)
- Optimistic UI updates: assets move between tabs instantly, with rollback on error. (The delete request on dev took awhile to respond, so it made the action feel sluggish without this). Pending items in trash are grayed out with disabled Restore button until server confirms.
- Table has global text filter and clickable column sorting
- Row removal and sorting animations
- Toast component adds action callback support (for `undo` action).
- Global QueryClient configured to skip retries on 401/403/404 errors. 
- Vibe-testing added and mock server updates to support ui changes.

On dev.


cmcfadden (commented):
Thoughts on adding a confirm on delete? Even though we have undelete, it's potentially expensive (since we have to regen derivatives)

jxjj:
Good call on the regen cost. How expensive is it on prod? Are we talking seconds or minutes (or longer)? Or do you mean "expensive" in terms of real aws compute $$$? Or both?

An earlier iteration had confirms, but I switched to toasts with "undo" after doing some asset cleanup locally and found that those confirm clicks really add up. 

A middle path is the "delayed send" pattern. User clicks delete, undo toast shows for a few seconds, request doesn't fire until the toast expires. Accidental deletes never hit the server, so no regen. 

Happy to go either way. If we do confirms, we should mention the est. regen time in the dialog so users know.

cmcfadden:
It'll vary depending on the asset(s) and spot pricing - A few minutes to many hours - but the folks doing deletes will used to that from uploading things in the first place. Probably most important for the undo toast to just say "digital assets will become available after derivatives are created" or something to that effect?

jxjj:
Changes:

## Delete / Restore Asset Warning
ScreenShot 2026-03-25 at 20 15 44@2x

Confirmation shown to users on first delete or restore, warning them that that derivatives will be removed/regenerated.

It only appears on first delete/restore, so users deleting many assets at once won't be repeatedly warned.

## Deleted asset notice (410 Gone handling)
ScreenShot 2026-03-25 at 20 18 25@2x

ScreenShot 2026-03-25 at 20 19 24@2x

Previously: when navigated to a deleted asset or an asset with a deleted related asset, the frontend should show a generic error toast and a loading spinner.

Now:
If a user has `canManageAssets` permissions:
- `410 Gone` is returned from server with enough context to restore asset, instead of `404`. (part of https://github.com/UMN-LATIS/elevator/pull/220). 
- frontend shows a contextual "Asset Deleted" notice with the deletion date instead of a generic error toast.
- Editor can use a `Restore` button to restore the asset inline.

For non-editors:
- if the asset is deleted they see a 404.
- if the asset has a deleted related asset, the asset will render correctly and not attempt to render the deleted related asset.

## Other fixes
- removed "undo" from toast action.
- fix delete animations firing when filtering
- fix issue where some deletes would not get a "success" toast (multiple deletes in flight would only show "success" for the last one)

On dev for another round of testing

cmcfadden (approved):
looks great

PR #498, opened 2026-03-31
add graceful handling when search totalResults and matches mismatch; prevent negative skeleton cards

### LINKED ISSUE #451: Error state when number of results returned doesn't match what we say there will be
If the number of results actually returned in the searchResults call is less than the number of of ids returned at the top, we can end up with this error state. In this case, there was an item that was deleted but which wasn't fully removed from the index, so there were 95 matches according to the summary but only actually 94 for real. 

Image

jxjj:
Resolves #451.

- Fix 1: When searching, it's possible that the backend's `totalResults` is inaccurate. We shouldn't use `matches < totalResults` to determine if there are more matches. Instead, we treat this as a placeholder for totalResults and then if we fetch a page with empty results we know we're done and update the ui accordingly. 
- Fix 2: Back navigation now correctly prevents resetting currentPage to 0 if we have more than 1 page loaded. This was causing issues where duplicate results would be loaded, leading `matches.length` to exceed `totalResults`.
- Fix 3: Related. If matches exceed total results, it was possible for the number of skeleton cards in the `v-for` to be negative. We now guard against negative counts.
- Also:`hasMoreResults` is now sourced from the store, not computed independently in each component so that other pages with search results (embedded search) will benefit from these fixes.

On dev.

https://dev.elevator.umn.edu/defaultinstance/search/s/33e71528-35e0-43ea-bdfb-4594d90d5c44?resultsView=grid

### DIFF
--- a/mock-server/db/searches.ts
+++ b/mock-server/db/searches.ts
@@ -48,7 +48,10 @@ export function createSearchesTable({
   return {
     ...baseTable,
     // Table-specific methods
-    create: (query: string): SearchResultsResponse => {
+    create: (
+      query: string,
+      { totalResultsOverride }: { totalResultsOverride?: number } = {}
+    ): SearchResultsResponse => {
       const searchId = crypto.randomUUID();
 
       // find any matches in the assets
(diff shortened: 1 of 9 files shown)


cmcfadden (approved):
this looks great thanks

PR #510, opened 2026-04-03
Fix preview image polling leak after leaving asset editor

### LINKED ISSUE #480: previewImage polling continues to run after leaving the asset creation page
Image

1. open asset without preview images in asset editor
2. previewImage polling starts
2. go back to "my assets"
3. observe previewImage polling continues

expected polling to stop when page unmounts

jxjj:
- Fixes #480
- Preview image polling now stops when navigating away from the asset editor, instead of continuing indefinitely
- Uses `watch` + `onBeforeUnmount` to unregister fileIds on unmount, with reference counting to handle shared fileIds safely


cmcfadden (approved):
cooooooool

PR #523, opened 2026-04-06
Fix login redirect when deep-linking to a protected asset

### LINKED ISSUE #522: login fails when linking to an asset and not logged in
hit this link in an incognito window:
https://dendro.elevator.umn.edu/asset/viewAsset/69bd6a445f97ea86e409b34a

then hit "guest login"

Image

jxjj:
- Fixes #522
- Unauthenticated users visiting a protected asset can now log in and get redirected back to the asset.
- Login notifcation appears directly on page, not in modal.
- Removes background 401 error.
- Add tests for the permissions issue.

Tricky bug.

When a 401 would happen, the page would look something like this.

```
<Page>
  <ErrorModal>
      <LoginNotification>
           <RouterLink :to="/login">Login</RouterLink>
      </LoginNotification>           
   </ErrorModal>
</Page>
```

Notice that the `<RouterLink>` button is **inside** the `<ErrorModal>`.

Then:
1. user clicks <RouterLink>
2. vue router invokes `beforeResolve()` hook
3. **beforeResolve() runs `errorStore.clearError()`** This is done so that errors don't persist across page navigations.
4. Vue does a DOM update. Now that the error is clear `<ErrorModal>` is torn down... and with it the `<LoginNotification>` and `<RouterLink>`.
5. After RouterLink is torn down, the route never resolves to the correct page.

###  Fixes

- We change to render the login notifcation directly on the asset page, not in the error modal to avoid the tear down. That should fix _this_ particular issue.
- BUT, I'm pretty sure other pages rely on the 401 modal (e.g. a user's auth expires). So, we also change from  `<RouterLink :to="...">` to `<Button @click="goToLocalLogin">`. The`goToLocalLogin()` does a `router.push()` to the page. Vue doesn't update the DOM until functions complete, so we're guaranteed that the redirect will resolve correctly before the component is torn down. We could have also cleared the Error on `nextTick()` I think.

### DIFF
--- a/mock-server/db/assets.ts
+++ b/mock-server/db/assets.ts
@@ -54,7 +54,7 @@ const baseAsset: WithMeta<Asset> = {
   titleObject: "title_1",
 };
 
-const generateMockAssets = (count = 100): Asset[] => {
+const generateMockAssets = (count = 100): WithMeta<Asset>[] => {
   const assets: Asset[] = [];
   const fileTypes = ["txt", "pdf", "docx", "jpg", "png", "mp4", "wav"];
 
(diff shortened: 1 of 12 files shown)


cmcfadden (approved):
seems good to me

PR #547, opened 2026-04-29
Show default OS scroll at top level, but keep styled scroll for child components

jxjj:
This removes scrollbar formatting so that Colin's scrollbars properly disappear when he's not using them. 😉

Whenever any `::webkit-scrollbar-*` styles are set in CSS, (some) browsers will disabled their own built-in styling and fall back to "classic" scrollbars. See: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

I _think_ the original design rationale was there's some features that may hide content unexpectedly if users have disappearing scrollbars. For example, the asset view side panel or search result cards with more content.

https://github.com/user-attachments/assets/61d83729-21da-403d-803e-6aa24acf606b

### DIFF
--- a/src/css/app.css
+++ b/src/css/app.css
@@ -63,22 +63,6 @@ body {
     clip-path: inset(50%);
     white-space: nowrap;
   }
-
-  ::-webkit-scrollbar {
-    width: 0.5rem;
-  }
-
-  ::-webkit-scrollbar-track {
-    background: var(--surface-container);
-  }
-
-  ::-webkit-scrollbar-thumb {
-    background: var(--outline);
-  }
-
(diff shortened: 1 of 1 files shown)


cmcfadden:
uggggggh you make a good point about the asset sidebars ...

what if we just hid them on the search page??

😭

jxjj:
(putting this here in case we forget why we made this decision)

If we do os native scroll on search page, we'd lose the scrollbars in the advanced search components, which isn't ideal. Instead this is updated to use os-default scroll at the top  `html` level, and then show scrollbar for any child components (css selector `html *`).

Issue #543, opened 2026-04-22
Sundry template editor improvements


jxjj:
Within the template editor:

- [ ]  duplicate button needs confirm dialog
- [ ]  remove reorder widgets button on index page
- [ ]  there's a lot of empty space on the left of the widgets, which could be better used -- especially when the options are options are open (e.g. json editor)
- [ ]  can't tell widget type is a select. Needs an outline.
- [ ]  show current template id someplace. it's needed for relationship fields.

When adding new widget within the editor:
   - [ ] options should be expanded by default
   - [ ] all settings should default to `off`
   - [ ] show internal field name (e.g. `images_1`) somehow. Permit users to edit it if needed. (but don't make it too easily to accidentally modify).

PR #549, opened 2026-05-01
template editor edits

jxjj:
Fixes several issues in #543:

- reworks the widget editor to increases the space available for editing widgets. Expand/collapse are consolidated into a single  icon button.
- adding a new widget starts with options expanded. this require a refactor since the expand/collapse all was just a hack before. Now we centrally track which widgets are expanded and collapsed. To do this, we needed to give a temporary stable id (indices won't work since widgets can be reordered). This adds a temp`_uuid` prop to each widget, which is stripped out on save.
- adds a templateId to the top of the page
- json textarea will start content sized (not yet baseline on firefox. Fallback is typical min-width)
- border added to widget type select box 

CleanShot 2026-04-30 at 18 57 24@2x

Now opens with options expanded by default:
https://github.com/user-attachments/assets/1608a97a-c324-42fb-be64-41eee1aeeb18


cmcfadden (approved):
looks great

PR #550, opened 2026-05-01
add field title as locked input

jxjj:
- Adds `<LockedInput>` which started an input field as readonly and locked until "Edit" is clicked. Change isn't committed (v-model isn't updated) until the change is confirmed by clicking the "check" button.
- Adds "Advanced" section Widget Options with the field title as locked input. Includes validation that it's set, only `a-z0-9_-` and unique to the template.

Stacked on #549 

https://github.com/user-attachments/assets/8e79e267-5748-4edb-93f1-b4c636273111


cmcfadden (approved):
looks good

PR #556, opened 2026-05-13
Fix/edit button on homepage

jxjj:
CleanShot 2026-05-13 at 13 22 35@2x

- fixes edit button
- adds script to remote deploy dev (taken from `develop` branch)

### DIFF
--- a/src/pages/StaticContentPage/StaticContentPage.vue
+++ b/src/pages/StaticContentPage/StaticContentPage.vue
@@ -47,7 +47,7 @@ const props = defineProps<{
   pageId: number;
 }>();
 
-const BASE_URL = config.instance.base.path;
+const BASE_URL = config.instance.base.url;
 const pageIdRef = toRef(props, "pageId");
 
 const canCurrentUserEdit = computed(() => {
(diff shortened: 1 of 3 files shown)


src/pages/StaticContentPage/StaticContentPage.vue
```diff
@@ -47,7 +47,7 @@ const props = defineProps<{
   pageId: number;
 }>();
 
-const BASE_URL = config.instance.base.path;
+const BASE_URL = config.instance.base.url;
```
cmcfadden, line 50: what's the story with this change? just wondering if this creates problems with our instances that don't run on domain model but instead subdirectories?

jxjj: `base.url === base.origin + base.path`

Prod instances with a base path should be unaffected unless there's some complication I'm not thinking of:
Before: `/defaultinstance`, AFTER: `https://dev.elevator.umn.edu/defaultinstance`

I made this change to test this feature locally by proxying to a backend server. With vite's proxy, the backend (dev.elevator.umn.edu) is mounted at `/api` so:
- BASE_ORIGIN=`/api` 
- BASE_PATH=`/defaultinstance`
- and so BASE_URL=`/api/defaultinstance`

Also: the variable is called BASE_URL, so this is more truthful.

cmcfadden (approved):

Issue #538, opened 2026-04-14
Implement admin / permissions in new ui


jxjj:
- Create New Group
	- Group type:
		- Add Another value
		- Standard
			- All users
			- Authenticated
			- Centrally Authed
		- Labelled (custom?)
			- Unit
			- Job Code
			- Class Number
			- Dept/Course Number
			- Student Status
			- Employee Type
			- Specific People
- Assigning Groups different Permission Levels
	- No Permissions
	- Search and Browse
	- View Derivatives (Group 1) -- filehandler group based on media?
	- View Derivatives (Group 1 & 2) 
	- Create/Edit Drawers
	- Download Originals
	- Add Assets to Instance (curator)
	- Admin
- My Users (Local User management)
	- Add/Edit a User
		- username
		- display name
		- password
		- repeat password
		- email
		- apikey (this should be generated, not typed)
		- api secret (same)
		- SuperAdmin?
		- Expires?

PR #557, opened 2026-05-22
feat: admin permissions foundation (#538, part 1)

jxjj:
CleanShot 2026-05-21 at 23 24 08@2x

Groundwork for the admin permissions UI (#538). First PR in a stack. Pure infrastructure, no permissions, groups, or rules UI yet.

I'm trying a lightweight feature-flag pattern across these PRs so `develop` stays always-shippable while each PR stays small(ish). New admin UI is gated behind `VITE_FEATURE_ADMIN_PERMISSIONS` (default off). A runtime override via `globalThis.Elevator.config` could also flip it per-instance without a rebuild if desired. With the flag off, this PR changes nothing user-visible.

- `<AdminLayout>` + `<AdminSidebar>`, plus `<PageContent>`/`<PageHeader>` page primitives
/admin/permissions route with a placeholder page (real page comes next)
- Menu: a flag-gated "Instance Permissions (WIP)" entry above the existing legacy link, which is unchanged

Try it locally: set `VITE_FEATURE_ADMIN_PERMISSIONS=true` in `.env`, restart Vite, click the new menu entry.


cmcfadden (approved):
change is fine with me - you can decide about the copilot stuff

Issue #561, opened 2026-06-02
Select Groups in new UI don't support key:value syntax


cmcfadden:
We support two ways to define select groups:

```
{
  "multiSelect": false,
  "selectGroup": [
    "option 1",
    "option 2",
    "option 3"
  ]
}
```

or:
```
{
  "multiSelect": false,
  "selectGroup": {
    "option 1": "more info about the option 1",
    "option 2": "more info about the option 2",
    "option 3", "more info about the option 3"
  }
}
```

It looks like the curator UI doesn't support the second one. Only the value shows in both the curator and visitor view. It should draw the key value (option 1, option 2, etc) for curators and the value for end users. It also looks like the new admin UI is only presenting the first syntax as an option in the sample it populates.

For an example template, check out "Audio Test" on dev.

jxjj:
Resolved-ish in #563

PR #562, opened 2026-06-02
proposed fix for #561

cmcfadden:
proposed fix for #561


jxjj:
Looks good to me. 

My questions are more UX and DX related:
- It seems like a curator wouldn't know what label a visitor would see without either saving and viewing the page, or inspecting the template. Is that an issue to address in the new UI or doesn't it come up?
- It might be beneficial to add a comment above the fix that it's intentional to show the curator the key and not the value and why. I can see coming across this in the future and wondering if it was a bug.

jxjj (approved):

PR #568, opened 2026-06-04
fix: search results map show NaN error when no results

### LINKED ISSUE #560: Map tab when you don't have locations breaks the app
going to the map tab when you don't have any locations in your results throws an error

Image

jxjj:
Resolves #560 

The issue occurred because v-show will render but not show the Map when there's no markers. `boundingBox` is lazily computed with `(Infinity, Infinity)` lnglats, and then computation of `center` becomes `(Infinity + Infinity)/2 = NaN`.

Changing to `v-else` so that the `<Map>` component isn't mounted if there are no markers resolves things.

(An alternative would be to change `getBoundingBox` to always return non-infinite values.)

on dev.

CleanShot 2026-06-04 at 15 20 41@2x

### DIFF
--- a/src/components/SearchResultsMap/SearchResultsMap.vue
+++ b/src/components/SearchResultsMap/SearchResultsMap.vue
@@ -14,7 +14,7 @@
       </Button>
     </div>
     <Map
-      v-show="markers.length > 0"
+      v-else
       :zoom="10"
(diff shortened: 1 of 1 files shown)


cmcfadden (approved):
thx

PR #569, opened 2026-06-06
handle file downloads that need to be thawed from Glacier

### LINKED ISSUE #288: Download "Give us a moment" is in the legacy interface
Image

### LINKED ISSUE #546: Downloads of files stored in Glacier currently download html
The "downloadOriginal" endpoint might return an HTML page for files that need to be restored from Glacier. The download original link in the new UI will save that file (as a tiff for example). 

This one will be a little annoying to test, because once you click a file and trigger a restore, it'll probably get restored and not be useful for testing again. 

Image

jxjj:
Resolves #546
Resolves #288 

When a user downloads an original, it's possible that the file needs to be restored from cold storage (Glacier) first. In the legacy ui, the user sees a "Give us a moment" message. The new ui would download the response from the server regardless, which would cause the html message to be saved as the file with the wrong file type and extension.

This PR handles the case better when downloading a single original or choosing "Download All Originals" from the asset.

Sketch of the fix:
1. When the user requests to download an original, the UI first probes the download endpoint to see what response type is returned.
2. If the response type is `text/html`, we assume that this isn't a download, and so we show the response text from the server. This handles both the "Give us a moment" case, and some error.
3. If the response is not `text/html` or a redirect, we proceed by downloading as normal.

When a user chooses "Download All Originals", if a text/html (or error) is return we show a toast, and skip to the next file rather than preventing all downloads until the archived file can be restored.

The fix is somewhat hacky. Arguably, a better approach would be to return the file storage status from a server endpoint, and have a separate endpoint for requesting the restore. This would save extra probes and permit showing clearer status to the user (e.g. that a restore is already in progress, or that it failed for some reason). But, given discussion of possible backend changes, the probe-first fix seemed good enough for now.

This adds a test asset to the mock server which behaves like it's been stored in Glacier. We don't test the "Download All" behavior.

Here's a video of both cases:

https://github.com/user-attachments/assets/874b1fce-2c76-410e-9ee2-34d87f3cdfd0


jxjj:
Discussed with @cmcfadden . Going for API approach, closing.

PR #570, opened 2026-06-16
feat: restore archived files from ui

### LINKED ISSUE #546: Downloads of files stored in Glacier currently download html

jxjj:
Resolves #546 

- Update `<DownloadFileButton>`. This now uses new endpoints to check the status of an original before attempting to download. If the original is archived, a "Restore Original" button is shown to the user. While restoring, the user sees a spinner and message. When restored, the user can download the original like normal.
- Update `<UploadWidget>` similarly so when the user chooses "Download All Originals", they will see toasts when: 1) download started, 2) restore started, 3) error.
- adds infra and tests for downloading originals


cmcfadden (approved):
well actually Elevator serves the whole University system so the goldy is problematic ...  ;)

PR #576, opened 2026-06-18
bump x size, add hover state

### LINKED ISSUE #572: Need a way to delete digital assets
If I've uploaded a file to an asset, but need to delete it and attach a new one, there's not currently a way to do that without deleting the whole record. Old UI has a way to delete an attachment.

jxjj:
Previously, the `X` to delete was a bit small and the hover effect was so subtle it could be missed. This bumps the icon size and changes hover to makes it clearer that the action is destructive.

If it's still too subtle, we could make the hover state the default state.

BEFORE
CleanShot 2026-06-18 at 15 20 33@2x

AFTER
CleanShot 2026-06-18 at 15 20 12@2x

ON HOVER
CleanShot 2026-06-18 at 15 24 26@2x

IN CONTEXT
CleanShot 2026-06-18 at 15 19 21@2x

Closes #572

### DIFF
--- a/src/pages/CreateOrEditAssetPage/EditWidget/EditWidgetLayout.vue
+++ b/src/pages/CreateOrEditAssetPage/EditWidget/EditWidgetLayout.vue
@@ -162,14 +162,14 @@
                         widgetDef.type === Types.WIDGET_TYPES.UPLOAD
                       "
                       :class="[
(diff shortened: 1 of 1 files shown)


cmcfadden (approved):
thank you for not just telling me I'm blind ;)

PR #577, opened 2026-06-26
add tab for admin group mangement

jxjj:
Adds part of the Groups UI slice for #538. Complements api  in https://github.com/UMN-LATIS/elevator/pull/238

This is gated behind a feature flag env variable. (Requires setting`VITE_FEATURE_ADMIN_PERMISSIONS=true` during the build phase to be available to users).

- Adds a working **Group** tab under Admin > Permissions (WIP). Instance admins can create, edit, and delete groups.
- Adds support for "Specific User" groups including member management: search, add, remove members.
- When a new group is created, the group name or member input will be focused in the UI.
- Support for value-based permission groups (e.g. course codes) aren't part of this slice. (The types will appear disabled in the group type dropdown).

Supporting changes:
- fixes Toast/Modal z-index conflict. Toast now correctly appear above standard modals. (e.g. error toast will appear above group creation modal).
- cancelled api requests are now properly handled without triggering global error (e.g. autocomplete sends updated query and cancels previous).
- TansStack query used for managing group requests, caching, invalidation, and request deduping.
- added `tryFocus` helper to better support focusing an input after update. It will retry (default 10 times) to find the dom element to focus before giving up. This handles an issue where Vue updates require more than one tick (so `await nextTick()` isn't enough).
- Some color fixes (inverse colors, modal surface color not different enough from input color)

on dev

### DIFF
--- a/bin/deploy-dev.sh
+++ b/bin/deploy-dev.sh
@@ -24,6 +24,14 @@ echo "→ Checking out $BRANCH..."
 git checkout "$BRANCH"
 git pull origin "$BRANCH"
 
+if [ ! -f .env ]; then
+  echo "→ Creating .env from .env.example..."
+  cp .env.example .env
+
+  echo "→ enabling admin permissions feature flag in .env..."
+  echo "VITE_FEATURE_ADMIN_PERMISSIONS=true" >> .env
+fi
+
 echo "→ Installing dependencies..."
 yarn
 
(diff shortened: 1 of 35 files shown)


cmcfadden (approved):
Because this is pretty chunky, I'm gonna just approve it and we'll plan an aggressive test cycle once we're at that point

PR #580, opened 2026-07-06
Add support for managing AuthHelper groups

jxjj:
Part of #538, slice 2 (auth-helper groups), UI side. Pairs with the elevator API PR.
- Enables all group types in the Create Group form. The "(not implemented)" disabled options are gone. 
- Adds an auth-helper value entries manager that's similar in form/function to the members manager: crud ops with confirm on removal.
- Similar to members, autocomplete for entries will suggest values on focus. The input always offers a verbatim typed text input.
- Feature still sits behind an env variable, so that develop remains deployable.

Other things:
- Add member/entry button moved to bottom of the table. Previously, the Add Member button was above the table. When clicked, the browse would focus the form, which may be off screen and cause a disorienting page scroll. Moreover, adding multiple members would require the user to scroll back up. This puts the button where the form will appear.
- Fixes `disabled` bug on `<InputGroup>` component. Previously, both the input and the group containing div would get a `disabled` attributed, causing some styling quirks. Now, only the `input` elements gets `disabled`.
 - Changed strategy for query invalidation. Now just invalidating `onSettled`, rather than doing the optimistic update and rollback for each query. This simplifies the reconciliation logic, and keeps the query data cache clean.
 - Also changed how query keys were assigned in TSQ. Previously one member or group addition, would invalidate group details for EVERY group, not just the one changed group and group list.
 - Consolidated all group tanstack queries into a single file: src/pages/AdminPermissionsPage/groupQueries.ts (which made the above changes easier).


cmcfadden (approved):
I think this is good - we can talk through some testing ideas when I'm back to make sure we're not at risk of any weird DB states

cmcfadden (approved):

PR #590, opened 2026-07-21
Add theme preview for admins

jxjj:
- Admins can preview any theme from Instance Settings: an eye button per theme applies it to the whole app without saving anything, and a bottom bar offers a theme switcher and Revert.
- The preview follows SPA navigation so admins can judge real pages (search, asset views, maps), while a reload or Revert restores the saved theme. Nothing is written to the API or localStorage.
- The Available Themes grid renders each theme as swatch card: the theme's own surface color, display font, and primary/secondary/tertiary dots, with the checkbox and preview button in the app theme beside it.

CleanShot 2026-07-21 at 11 01 42@2x

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -13,6 +13,7 @@
     <Teleport to="body">
       <ErrorModal />
       <ToastRoot class="z-[60]" />
+      <ThemePreviewPill />
     </Teleport>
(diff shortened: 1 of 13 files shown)


cmcfadden (approved):
looks great

PR #599, opened 2026-07-23
make it easier to set mock server port in dev

jxjj:
This permits more than one mock server to run similtaneously (which is friendlier to worktrees). 
- adds an env variable to `.env.example`
- uses node's newish feature to read in the MOCK_SERVER_PORT variable from .env

Impacts dev/test environment only


cmcfadden (approved):
overachiever

PR #600, opened 2026-07-24
Surface file descriptions in File Info and treat hidden descriptions as alt text

### LINKED ISSUE #416: Surface description in info window?
Now that we've got these nice AI generated descriptions, we should think about somewhere to surface them for the user? And then maybe rethink the truncated descriptions below them? Can brainstorm a bit

jxjj:
- Resolves #416
- File descriptions now always appear in the "More Info" (File Info) modal, and the "Show Description below Thumbnails" instance setting is now a real mode switch: descriptions shown to everyone and used as the accessible name when on, image alt text only when off.
- When descriptions are visible, thumbnail images use an empty alt so screen readers announce the text once (the W3C functional-images pattern, linked in code comments). When hidden, the description becomes the alt and no tooltip renders.
- The asset editor's field label follows the mode ("Description / Alt Text" vs "Alt Text"), and the instance setting gains a help "?" icon explaining the switch.

Instance Setting:
CleanShot 2026-07-24 at 11 02 41@2x

When showing:
CleanShot 2026-07-24 at 11 06 18@2x
CleanShot 2026-07-24 at 11 04 10@2x

When off:
CleanShot 2026-07-24 at 11 01 58@2x
CleanShot 2026-07-24 at 11 07 18@2x

Always shows description/alt in File Info:
CleanShot 2026-07-24 at 11 03 56@2x


cmcfadden (approved):
rad!

PR #601, opened 2026-07-24
add `addOnPaste` prop to break pasted tags on `,`

### LINKED ISSUE #559: Allow pasting comma separated values to a tags field
if you paste a string like:
test, test2, test3

into a tag field, it gets considered a single term instead of splitting apart. We should probably make it more consistent by parsing it into tags?

jxjj:
- Resolves #559 
- breaks tags apart on comma when pasted

https://github.com/user-attachments/assets/5b48c2e4-ce8a-417f-ab99-591ba880fa56


cmcfadden (commented):
nice

cmcfadden (approved):

PR #641, opened 2026-07-30
fix asset editor data bugs

jxjj:
> Note: most of the added lines in this PR are tests. The production change is 10 files and about **470 net lines**, and 236 of the deletions are one file being renamed. If you read two things, read `editorReducer.ts` and the `useAssetEditor.ts` diff.

This fixes several subtle asset editor data-loss bugs that Claude found in an audit (which started with just trying to get to the root cause of typescript complaint).

- Replaces the editor's loose collection of refs with **one state type** and **one pure function** that moves between states, so every way the editor can change is written down in one place.
- Fixes **bugs on `develop` that could lose data**. Four are save races that could write to the wrong asset. Three involve inline related assets: a failed save reported as a success, writes to an asset the user already left, and an orphan created from an abandoned draft. All bugs have tests which replicate and confirm fix.
- Changes how the asset draft is stored. Instead of a second full copy of the asset, we just track the diff to the saved version (and the saved version itself). That deletes the dirty-checking machinery rather than fixing it.
- Takes the editor's core from 1 test to 45, none of them requiring mocks, plus three new end-to-end tests.
- The original typescript error that started this whole thing resolved naturally as a consequence of the design.


jxjj:
Superseded by #660

PR #646, opened 2026-07-31
Fix flakey drag and drop in template editor

### LINKED ISSUE #613: Template Editor: Drag and drop field ordering can be flakey. Keyboard is fine.
https://github.com/user-attachments/assets/f50fa096-31e8-4035-93fc-bff39c05675d

jxjj:
- Fixes #613 
- Dragging a field to a new position now takes, in Chrome and Safari. It used to snap back to where it started.
- Root cause: all three drag and drop components built a list of cleanup functions but called `combine` on it at setup time, while it was still empty, so the unmount hook was a no-op. A remounted list left a second drop monitor alive, both handled the drop, and applying one reorder twice returns the item to its starting position.
- Also fixes three smaller problems in the same component: no focus ring on the drag handle in Safari, a drop line that jumped between two positions in each gap, and a scrollbar that flashed for the length of every drag.

https://github.com/user-attachments/assets/fd139280-0d7d-4daf-826e-298e9096aabf

### DIFF
--- a/src/components/DragDropList/DragDropContainer.vue
+++ b/src/components/DragDropList/DragDropContainer.vue
@@ -33,13 +33,9 @@ const emit = defineEmits<{
 
 provide(GROUP_ID_PROVIDE_KEY, props.groupId);
 
-type CleanupFn = () => void;
-const cleanupFns = [] as CleanupFn[];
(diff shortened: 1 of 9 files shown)


cmcfadden (approved):
hoorah!

